Issue #5, opened 2022-06-27
[META] Rule curation layer(P0) & Rule ingestion layer(P0)


sbcd90:
**Is your feature request related to a problem?**

One of the important features of a Security Analytics solution is to define a mechanism to ingest & store security related rules into it. As of today, Sigma provides a comprehensive set of security related rules which can be used to identify threats. Sigma also provides a well-defined generic schema for defining new Sigma rules.

In order to reuse the existing set of Sigma rules & to support the existing Sigma schema, a Rule curation layer is required, which transforms, converts & aggregates Sigma rules into Opensearch queries. A Rule ingestion layer is required to ingest these queries into Opensearch.

**What solution would you like?**
There are 2 approaches considered for designing the Rule curation layer.

- A separate Opensearch specific Sigma backend converter

In this approach, the user needs to manually run the Opensearch backend converter to convert the rule yaml file defined using Sigma schema into Opensearch query & provide it as input to the plugin.

- A built-in Sigma rule parser

In this approach, the user just provides the rule yaml file defined using Sigma schema & the plugin takes care of converting the rule yaml into Opensearch query & finally ingest these queries into Opensearch.
This approach is especially beneficial for defining ad-hoc Sigma rules & allows users without Opensearch query syntax knowledge to also use this tool.
The `P1 Item Rule Editor Framework` is not possible without this approach.

**Do you have any additional context?**

The overall scope of MVP is defined [Security Analytics in OpenSearch (MVP)](https://github.com/opensearch-project/security-analytics/issues/7)

**Tasks**

- [X] integrate sigma rules-types, exeptions, utils
- [ ] integrate sigma rules-modifiers, conditions
- [ ] integrate sigma rules-objects, query backends

PR #26, opened 2022-08-16
add all sigma rules for security analytics

sbcd90:
Signed-off-by: Subhobrata Dey <sbcd90@gmail.com>

### Description
this PR adds all pre-packaged sigma rules into the security-analytics repo.

### Issues Resolved
[Rule ingestion layer](https://github.com/opensearch-project/security-analytics/issues/5)
 
### Check List
  - [ ] All tests pass
- [X] New functionality has been documented.


src/main/resources/OSMapping/fieldmappings.yml
```diff
@@ -0,0 +1,5 @@
+fieldmappings:
+    EventID: event_uid
+    HiveName: unmapped.HiveName
+    fieldB: mappedB
+    fieldA1: mappedA
```
getsaurabh02, line 5: Looks like some mapping template. Can we add comments to suggest this is a reference template for new templates to be added in future.

getsaurabh02 (changes requested):
Thanks @sbcd90. I am wondering how are we going to categorize these rules and map them   with the log type, so that we can associate detectors with the right set of rules. 

I see one of the possible approaches as we pre-bucketize these rules into specific paths, based on the log types they are useful for, and then we and a mapping file which maps the Rule Paths (folder/subfolder/specific-rule) to the 8 log groups that we will support initially. 

As we will add new log group type, we can add more rules and update the mapping. Let me know if you concur with the thought or have any alternative proposal.

src/main/resources/OSMapping/fieldmappings.yml
sbcd90: fixed it & added comments to yaml files.

getsaurabh02 (changes requested):

src/main/java/org/opensearch/securityanalytics/rules/backend/QueryBackend.java
```diff
@@ -57,7 +57,7 @@ public QueryBackend(boolean convertAndAsIn, boolean enableFieldMappings, boolean
         this.errors = new ArrayList<>();
 
         if (this.enableFieldMappings) {
-            InputStream is = this.getClass().getClassLoader().getResourceAsStream("OSMapping/fieldmappings.yml");
+            InputStream is = this.getClass().getClassLoader().getResourceAsStream("OSMapping/windows/fieldmappings.yml");
```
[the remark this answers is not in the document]
sbcd90: @getsaurabh02 this will be made dynamic in the 3rd pr. the static mapping is kept here just for test cases to pass for this pr.

getsaurabh02 (approved):

eirsep (approved):

PR #32, opened 2022-09-10
adds create detector api flow

sbcd90:
Signed-off-by: Subhobrata Dey <sbcd90@gmail.com>

### Description
 - adds create detector api flow. author: @sbcd90 
 - adds support for custom datasources in `security-analytics`. author: @eirsep 
 
### Check List
  - [ ] All tests pass


eirsep:
I have a concern i wanted to call out with regards to the use of index document models in request layer which i have noticed in both Alerting, SAP i.e. usage of Detector in IndexDetectorRequest..

this doesnt seem to be a right practive.. 

I feel it would be better to have a Detector DTO at rest layer which should be used to build or which can be built from Detector document stored in the index.

This ensures fields of detector which dont need to be exposed are abstracted away from rest responses and transport level responses(if required)

@getsaurabh02 @sbcd90 what are your thoughts on this.

src/main/java/org/opensearch/securityanalytics/model/Detector.java
```diff
@@ -0,0 +1,439 @@
...
+        CLOUD,
+        COMPLIANCE,
+        LINUX,
+        MACOS,
+        NETWORK,
+        PROXY,
+        WEB,
+        WINDOWS;
+
+        public String getDetectorType() throws IOException {
```
eirsep, line 174: Enum allows properties which is neater and avoids switch case by implicit and mandatory declaration and mainly avoids unnecessary Exception handling.

```
public enum DetectorType {
        APPLICATION("application"),
        APT("apt"),
        CLOUD("cloud"),
        COMPLIANCE("compliance"),
        LINUX("linux"),
        MACOS("macos"),
        NETWORK("network"),
        PROXY("proxy"),
        WEB("web"),
        WINDOWS("windows");

        public String getDetectorType() {
            return type;
        }

        private String type;

        DetectorType(String type) {
            this.type = type;
        }
    }
```

src/main/java/org/opensearch/securityanalytics/model/Detector.java
```diff
@@ -0,0 +1,439 @@
...
+                    List<String> allowedTypes = Arrays.stream(DetectorType.values()).map(it -> {
+                        try {
+                            return it.getDetectorType();
+                        } catch (IOException e) {
+                            return null;
+                        }
+                    }).collect(Collectors.toList());
+
+                    if (!allowedTypes.contains(detectorType.toLowerCase(Locale.ROOT))) {
+                        throw new IllegalStateException("Monitor type should be one of ");
```
eirsep, line 287: plz add list of detector types

eirsep: illegalArgumentException would be more suited here?

src/main/java/org/opensearch/securityanalytics/action/IndexDetectorRequest.java
```diff
@@ -0,0 +1,74 @@
...
+    private WriteRequest.RefreshPolicy refreshPolicy;
+
+    private RestRequest.Method method;
+
+    private Detector detector;
+
+    public IndexDetectorRequest(
+            String detectorId,
+            WriteRequest.RefreshPolicy refreshPolicy,
+            RestRequest.Method method,
```
eirsep, line 30: don't think its right to use RestRequest.Method here. This object should be limited to rest layer if possible. 

maybe just use some enum Action{CREATE, UPDATE} or a boolean create

src/main/java/org/opensearch/securityanalytics/action/IndexDetectorAction.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.securityanalytics.action;
+
+import org.opensearch.action.ActionType;
+
+public class IndexDetectorAction extends ActionType<IndexDetectorResponse> {
```
eirsep, line 9: Is this a practice followed in opensearch?

can we have CreateDetectorAction and UpdateDetectorAction and corresponding request, response, transport and rest layer?

eirsep (changes requested):
Thanks for these changes Subho

i have some initial comments reg. structuring which I would like to discuss to understand best practices which we can address right away as we are writing new logic

src/main/java/org/opensearch/securityanalytics/model/Detector.java
sbcd90: fixed it.

src/main/java/org/opensearch/securityanalytics/model/Detector.java
sbcd90: fixed it.

src/main/java/org/opensearch/securityanalytics/action/IndexDetectorResponse.java
```diff
@@ -0,0 +1,87 @@
...
+            .field(_VERSION, version);
+        builder.startObject("detector")
+            .field(Detector.NAME_FIELD, detector.getName())
+            .field(Detector.DETECTOR_TYPE_FIELD, detector.getDetectorType())
+            .field(Detector.ENABLED_FIELD, detector.getEnabled())
+            .field(Detector.SCHEDULE_FIELD, detector.getSchedule())
+            .field(Detector.INPUTS_FIELD, detector.getInputs())
+            .field(Detector.LAST_UPDATE_TIME_FIELD, detector.getLastUpdateTime())
+            .field(Detector.ENABLED_TIME_FIELD, detector.getEnabledTime())
+            .field(Detector.ALERTING_MONITOR_ID, detector.getMonitorId())
```
getsaurabh02, line 67: looks much cleaner, thanks.

src/main/java/org/opensearch/securityanalytics/config/monitors/DetectorMonitorConfig.java
```diff
@@ -0,0 +1,89 @@
...
+import org.opensearch.securityanalytics.model.Detector;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+
+public class DetectorMonitorConfig {
```
getsaurabh02, line 16: These changes looks to be coming from a different PR. Are we merging these as well?

sbcd90: yes. they are part of the commit from @eirsep . https://github.com/opensearch-project/security-analytics/pull/32/commits/69f70a48e81b0ce2f89f0ad1dc2243cc6b5f79a
this is needed in the pr because otherwise our integration tests will fail.  https://github.com/opensearch-project/security-analytics/blob/3618a2056fee66e30daae1cb47bfa2dae720ea5f/src/test/java/org/opensearch/securityanalytics/resthandler/DetectorRestApiIT.java#L25-#L52

getsaurabh02 (dismissed):

getsaurabh02 (approved):
Thanks @sbcd90 Lets add the list of independent PRs we are trying to merge here in the description.

eirsep (approved):

PR #48, opened 2022-10-04
adds support for list of monitors in detector

eirsep:
Signed-off-by: Surya Sashank Nistala <snistala@amazon.com>

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/model/Detector.java
+++ b/src/main/java/org/opensearch/securityanalytics/model/Detector.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
@@ -81,7 +82,7 @@ public class Detector implements Writeable, ToXContentObject {
 
     private List<DetectorInput> inputs;
 
-    private String monitorId;
+    private List<String> monitorIds;
 
     private String ruleIndex;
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/securityanalytics/model/Detector.java
```diff
@@ -355,20 +361,20 @@ public static Detector parse(XContentParser xcp, String id, Long version) throws
...
+                version,
+                Objects.requireNonNull(name, "Detector name is null"),
+                enabled,
+                Objects.requireNonNull(schedule, "Detector schedule is null"),
+                lastUpdateTime != null ? lastUpdateTime : Instant.now(),
+                enabledTime,
+                DetectorType.valueOf(detectorType.toUpperCase(Locale.ROOT)),
+                user,
+                inputs,
+                monitorIds,
```
getsaurabh02, line 374: Do we also need Monitor Versions here?

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorRestApiIT.java
```diff
@@ -50,7 +50,7 @@ public void testCreatingADetector() throws IOException {
         Assert.assertFalse(((Map<String, Object>) responseBody.get("detector")).containsKey("findings_index"));
         Assert.assertFalse(((Map<String, Object>) responseBody.get("detector")).containsKey("alert_index"));
 
-        String monitorId = ((Map<String, Object>) responseBody.get("detector")).get("monitor_id").toString();
+        String monitorId = ((List<String>) ((Map<String, Object>) responseBody.get("detector")).get("monitor_id")).get(0);
```
getsaurabh02, line 53: Lets create a test which indexes multiple monitors and verify detector response.

src/main/java/org/opensearch/securityanalytics/model/Detector.java
eirsep: We won't need monitor versions as we can fetch monitors just using id.

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorRestApiIT.java
eirsep: added response parsing tests
Created issue #51 to track indexing multiple monitors.

Currently only added support for handling list of monitors in detector model and response.

sbcd90 (approved):

getsaurabh02 (approved):

PR #80, opened 2022-10-26
Rule field validation during IndexRule Action

### LINKED ISSUE #90: [FEATURE] API to validate list of rules against an index
We need a way to validate list of custom rules against source index so that user can be informed if specific rule can be used with particular index.

Create new API ValidateRules which accepts list of rule ids of same category(for example: "windows") and index name. Result will contain list of rule ids of rules which are non-applicable to given index. These rules contain fields which are not present in either given index mappings or alias mappings for this rule category.

REQUEST:

```
GET /_plugins/_security_analytics/rules/validate
{
    "rules": [
        "ruleId1",
        "ruleId2",
        "ruleId3"
    ],
    "index_name": "my_win2022_index_log"
}
```
Request body:
| param | description |
| --- | --- |
| **_rules_** | List of rule ids to validate |
| **_index_name_** | Name of index to validate rules against |

RESPONSE:

```
{
    "nonapplicable_rules": [
        "ruleId1",
        "ruleId3"    
    ]
}
```
| param | description |
| --- | --- |
| **_invalid_rules_** | List of rule ids which are not applicable to given index |

Validating rule during rule indexing.

petardz:
### Description
Added new API ValidateRules which, for a given indexName and list of ruleIds, will return list of non-applicable ruleIds to given index
 
### Issues Resolved
#90
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/securityanalytics/transport/TransportIndexRuleAction.java
```diff
@@ -181,8 +188,35 @@ void prepareRuleIndexing() {
                 final QueryBackend backend = new OSQueryBackend(category, true, true);
                 List<Object> queries = backend.convertRule(parsedRule);
 
-                Rule ruleDoc = new Rule(NO_ID, NO_VERSION, parsedRule, category, queries.stream().map(Object::toString).collect(Collectors.toList()), rule);
-                indexRule(ruleDoc);
+                //verify rule
+                verifyRule(backend.getQueryFields().keySet(), DetectorMonitorConfig.getRuleIndex(category), new ActionListener<>() {
```
getsaurabh02, line 192: What if the custom rule is created using the fields not part of the rule alias mapping, but using some additional fields from the Source Index. That still is a valid use case. 
Should we revisit the approach, where the fields in rule could also be present in the source index mapping?

getsaurabh02 (changes requested):

getsaurabh02:
Hi @petardz are we on it with the plan as we discussed ?

getsaurabh02 (dismissed):

eirsep (approved):

sbcd90 (approved):

PR #221, opened 2023-01-04
Flaky test fix: Execute doc monitor to ensure that doc level query index is created and populated

eirsep:
`org.opensearch.securityanalytics.resthandler.DetectorMonitorRestApiIT.testReplaceAggregationRuleWithDocRule_verifyFindings_success` is currently flaky.

Verification of query index fails intermittently. Added execute monitor command to ensure that query index exists.

Signed-off-by: Surya Sashank Nistala <snistala@amazon.com>


getsaurabh02 (commented):
Not sure why are we adding zips here

Issue #207, opened 2022-12-21
[BUG] Validate for existing name when creating detector


amsiglan:
**What is the bug?**
When user creates a detector, we are not verifying if a detector with same name already exists.

**How can one reproduce the bug?**
1. Create a detector using create detector API
2. Create another detector with same name

**What is the expected behavior?**
API fails to created the second detector with error along the lines of detector with same name already exists

**What is your host/environment?**
NA

**Do you have any screenshots?**
Na

**Do you have any additional context?**
NA

sbcd90:
@raj-chak assigned to you.

PR #261, opened 2023-01-10
changes to prevent creating two detectors with same name

raj-chak:
Signed-off-by: Raj Chakravarthi <raj@icedome.ca>

### Description
changes to prevent creating two detectors with same name
 Updated PR with a new approach. This approach uses name as the id for detector and onus is not on elasticsearch to generate id for detector after this change.
### Issues Resolved
(https://github.com/opensearch-project/security-analytics/issues/207)
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -686,6 +694,58 @@ void prepareDetectorIndexing() throws IOException {
 
         void createDetector() {
             Detector detector = request.getDetector();
+
+            //if  (detectorIndices.detectorIndexExists()) {
+            if  (false) {
```
getsaurabh02, line 699: This condition looks weird. Can we fix it

getsaurabh02 (approved):

eirsep:
will this work if 2 concurrent requests are made ?

praveensameneni:
@sbcd90 , @jowg-amazon , can you please review as we need this change - should not have detectors with same name as it will be a confusing experience.

Issue #226, opened 2023-01-04
Add field to store composite monitor id in detector document


eirsep:

stevanbz:
The PR that is created for parent issue containing all the changes:
https://github.com/opensearch-project/security-analytics/pull/316

engechas:
Closing as completed by #394

Issue #222, opened 2023-01-04
Update composite monitor during detector updation


eirsep:
When detector is updated, if new monitors are created and/or old monitors are deleted we also need to update the composite monitor sequence.
Update detector flow would involve the following interactions with alerting:
- create new monitors if any
- update sequence of detector's composite monitor by removing old monitor ids(if any monitors need to be deleted) and add new monitor ids
- if step one or two fail delete any newly created monitors, else delete monitors which are not required anymore

stevanbz:
Followed the steps from the above. The PR that is created for parent issue containing all the changes:
https://github.com/opensearch-project/security-analytics/pull/316

engechas:
This functionality is implemented, closing

Issue #220, opened 2023-01-04
Create composite monitor during detector creation


eirsep:
Composite monitors run a set of monitors in a sequence. Each detector should create bucket level and doc level monitors as required in a disabled mode and create a composite monitor workflow with a sequence of the above created monitor ids. They can optionally use the findings of one monitor to filter data queried by the next monitor.

- detector creation would involve following interactions with alerting:
 i. create doc level and bucket level monitors as per detector rules in disabled mode.
 ii. create composite monitors with sequence of above created monitor ids.
 iii. populate chained findings field to enable correct creation of detector triggers.
iv. delete all resources if `step i` or `step ii` fail.

stevanbz:
I am working on this. The idea is to create workflow for new detectors and follow the steps @eirsep mentioned in description.

stevanbz:
Followed the steps from the above. The PR that is created for parent issue containing all the changes:
https://github.com/opensearch-project/security-analytics/pull/316

getsaurabh02:
Removing untriaged as this is WIP feature. Lets follow the same for new feature issues @eirsep

engechas:
This functionality is implemented. Closing

PR #394, opened 2023-04-10
Extended detector use-cases to consider the workflow

stevanbz:
### Description
This PR contains the changes related with workflow. 
1. Enables workflow creation - all the monitors related with the given detector will be part of the detector ([220](https://github.com/opensearch-project/security-analytics/issues/220))
2. Enables workflow update during detector update ([222](https://github.com/opensearch-project/security-analytics/issues/222))
3. Deletes the workflow during the detector deletion ([225](https://github.com/opensearch-project/security-analytics/issues/222))
4. Adds additional field to a detector ([226](https://github.com/opensearch-project/security-analytics/issues/226))

This PR considers if the current detectors are supporting the workflow or not
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/model/Detector.java
+++ b/src/main/java/org/opensearch/securityanalytics/model/Detector.java
@@ -51,6 +51,8 @@ public class Detector implements Writeable, ToXContentObject {
     public static final String ENABLED_TIME_FIELD = "enabled_time";
     public static final String ALERTING_MONITOR_ID = "monitor_id";
 
+    public static final String ALERTING_WORKFLOW_ID = "workflow_ids";
+
     public static final String BUCKET_MONITOR_ID_RULE_ID = "bucket_monitor_id_rule_id";
     private static final String RULE_TOPIC_INDEX = "rule_topic_index";
 
@@ -98,6 +100,8 @@ public class Detector implements Writeable, ToXContentObject {
 
     private Map<String, String> ruleIdMonitorIdMap;
 
+    private List<String> workflowIds;
+
     private String ruleIndex;
 
     private String alertsIndex;
@@ -116,7 +120,7 @@ public class Detector implements Writeable, ToXContentObject {
                     Instant lastUpdateTime, Instant enabledTime, DetectorType detectorType,
                     User user, List<DetectorInput> inputs, List<DetectorTrigger> triggers, List<String> monitorIds,
                     String ruleIndex, String alertsIndex, String alertsHistoryIndex, String alertsHistoryIndexPattern,
-                    String findingsIndex, String findingsIndexPattern, Map<String, String> rulePerMonitor) {
+                    String findingsIndex, String findingsIndexPattern, Map<String, String> rulePerMonitor, List<String> workflowIds) {
         this.type = DETECTOR_TYPE;
 
         this.id = id != null ? id : NO_ID;
@@ -138,6 +142,7 @@ public class Detector implements Writeable, ToXContentObject {
         this.findingsIndex = findingsIndex;
         this.findingsIndexPattern = findingsIndexPattern;
         this.ruleIdMonitorIdMap = rulePerMonitor;
+        this.workflowIds = workflowIds != null ? workflowIds : Collections.emptyList();
 
         if (enabled) {
             Objects.requireNonNull(enabledTime);
@@ -164,7 +169,8 @@ public class Detector implements Writeable, ToXContentObject {
                 sin.readString(),
                 sin.readString(),
                 sin.readString(),
-                sin.readMap(StreamInput::readString, StreamInput::readString)
+                sin.readMap(StreamInput::readString, StreamInput::readString),
+                sin.readStringList()
             );
     }
 
@@ -199,6 +205,10 @@ public class Detector implements Writeable, ToXContentObject {
         out.writeString(ruleIndex);
 
         out.writeMap(ruleIdMonitorIdMap, StreamOutput::writeString, StreamOutput::writeString);
+
+        if (workflowIds != null) {
+            out.writeStringCollection(workflowIds);
+        }
     }
 
     public XContentBuilder toXContentWithUser(XContentBuilder builder, Params params) throws IOException {
@@ -287,6 +297,14 @@ public class Detector implements Writeable, ToXContentObject {
         }
 
         builder.field(ALERTING_MONITOR_ID, monitorIds);
+
+        if (workflowIds == null) {
+            builder.nullField(ALERTING_WORKFLOW_ID);
+        } else {
+            builder.field(ALERTING_WORKFLOW_ID, workflowIds);
+        }
+
+
         builder.field(BUCKET_MONITOR_ID_RULE_ID, ruleIdMonitorIdMap);
         builder.field(RULE_TOPIC_INDEX, ruleIndex);
         builder.field(ALERTS_INDEX, alertsIndex);
@@ -332,6 +350,7 @@ public class Detector implements Writeable, ToXContentObject {
         List<DetectorInput> inputs = new ArrayList<>();
         List<DetectorTrigger> triggers = new ArrayList<>();
         List<String> monitorIds = new ArrayList<>();
+        List<String> workflowIds = new ArrayList<>();
         Map<String, String> rulePerMonitor = new HashMap<>();
 
         String ruleIndex = null;
@@ -412,6 +431,13 @@ public class Detector implements Writeable, ToXContentObject {
                         monitorIds.add(monitorId);
                     }
                     break;
+                case ALERTING_WORKFLOW_ID:
+                    XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_ARRAY, xcp.currentToken(), xcp);
+                    while (xcp.nextToken() != XContentParser.Token.END_ARRAY) {
+                        String workflowId = xcp.text();
+                        workflowIds.add(workflowId);
+                    }
+                    break;
                 case BUCKET_MONITOR_ID_RULE_ID:
                     rulePerMonitor= xcp.mapStrings();
                     break;
@@ -463,7 +489,9 @@ public class Detector implements Writeable, ToXContentObject {
                 alertsHistoryIndexPattern,
                 findingsIndex,
                 findingsIndexPattern,
-                rulePerMonitor);
+                rulePerMonitor,
+                workflowIds
+            );
     }
 
     public static Detector readFrom(StreamInput sin) throws IOException {
@@ -599,8 +627,20 @@ public class Detector implements Writeable, ToXContentObject {
         this.ruleIdMonitorIdMap = ruleIdMonitorIdMap;
     }
 
+    public void setWorkflowIds(List<String> workflowIds) {
+        this.workflowIds = workflowIds;
+    }
+
+    public List<String> getWorkflowIds() {
+        return workflowIds;
+    }
+
     public String getDocLevelMonitorId() {
         return ruleIdMonitorIdMap.get(DOC_LEVEL_MONITOR);
+    }
+
+    public boolean isWorkflowSupported() {
+        return workflowIds != null && !workflowIds.isEmpty();
     }
 
     @Override
--- a/src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java
+++ b/src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java
@@ -32,6 +32,7 @@ import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.commons.alerting.AlertingPluginInterface;
 import org.opensearch.commons.alerting.action.DeleteMonitorRequest;
 import org.opensearch.commons.alerting.action.DeleteMonitorResponse;
+import org.opensearch.commons.alerting.action.DeleteWorkflowResponse;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.rest.RestStatus;
@@ -40,8 +41,10 @@ import org.opensearch.securityanalytics.action.DeleteDetectorRequest;
 import org.opensearch.securityanalytics.action.DeleteDetectorResponse;
 import org.opensearch.securityanalytics.mapper.IndexTemplateManager;
 import org.opensearch.securityanalytics.model.Detector;
+import org.opensearch.securityanalytics.util.MonitorService;
 import org.opensearch.securityanalytics.util.RuleTopicIndices;
 import org.opensearch.securityanalytics.util.SecurityAnalyticsException;
+import org.opensearch.securityanalytics.util.WorkflowService;
 import org.opensearch.tasks.Task;
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.TransportService;
@@ -59,6 +62,10 @@ public class TransportDeleteDetectorAction extends HandledTransportAction<Delete
 
     private final NamedXContentRegistry xContentRegistry;
 
+    private final WorkflowService workflowService;
+
+    private final MonitorService monitorService;
+
     private final ThreadPool threadPool;
 
     private final IndexTemplateManager indexTemplateManager;
@@ -71,6 +78,9 @@ public class TransportDeleteDetectorAction extends HandledTransportAction<Delete
         this.xContentRegistry = xContentRegistry;
         this.threadPool = client.threadPool();
         this.indexTemplateManager = indexTemplateManager;
+        this.monitorService = new MonitorService(client);
+        this.workflowService = new WorkflowService(client, monitorService);
+
     }
 
     @Override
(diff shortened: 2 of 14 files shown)


src/main/java/org/opensearch/securityanalytics/model/Detector.java
```diff
@@ -51,6 +51,8 @@ public class Detector implements Writeable, ToXContentObject {
     public static final String ENABLED_TIME_FIELD = "enabled_time";
     public static final String ALERTING_MONITOR_ID = "monitor_id";
 
+    public static final String ALERTING_WORKFLOW_ID = "workflow_ids";
```
eirsep, line 54: NIT: shouldnt we have single workflow id?
Any motivations for list?

src/main/java/org/opensearch/securityanalytics/model/Detector.java
```diff
@@ -412,6 +431,13 @@ public static Detector parse(XContentParser xcp, String id, Long version) throws
                         monitorIds.add(monitorId);
                     }
                     break;
+                case ALERTING_WORKFLOW_ID:
+                    XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_ARRAY, xcp.currentToken(), xcp);
+                    while (xcp.nextToken() != XContentParser.Token.END_ARRAY) {
+                        String workflowId = xcp.text();
```
eirsep, line 437: null check?

eirsep (changes requested):
what happens 
1. User creates detector in 2.6
2. upgrades cluster to version where workflows are introduced
3. updates detector rules? Will new workflow be created and the monitors be disabled? or will we not create workflows for existing detectors.
4. let's closely check the udpate scenario

src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java
```diff
@@ -137,35 +147,57 @@ public void onFailure(Exception t) {
...
-                @Override
-                public void onResponse(Collection<DeleteMonitorResponse> responses) {
-                    SetOnce<RestStatus> errorStatusSupplier = new SetOnce<>();
-                    if (responses.stream().filter(response -> {
-                        if (response.getStatus() != RestStatus.OK) {
-                            log.error("Detector not being deleted because monitor [{}] could not be deleted. Status [{}]", response.getId(), response.getStatus());
-                            errorStatusSupplier.trySet(response.getStatus());
-                            return true;
+            // If detector doesn't have the workflows it means that older version of the plugin is used
+            if (detector.isWorkflowSupported()) {
```
eirsep, line 151: debug log "deleting workflow <id> before deleting detector<id>

src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java
```diff
@@ -137,35 +147,57 @@ public void onFailure(Exception t) {
...
+                // 1. Delete workflow
+                workflowService.deleteWorkflow(detector.getWorkflowIds().get(0),
+                    new ActionListener<>() {
+                        @Override
+                        public void onResponse(DeleteWorkflowResponse deleteWorkflowResponse) {
```
eirsep, line 156: debug log "workflow <id> deleted. deleting monitors before detector deletion<id>"

src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java
```diff
@@ -137,35 +147,57 @@ public void onFailure(Exception t) {
...
-            ActionListener<DeleteMonitorResponse> deletesListener = new GroupedActionListener<>(new ActionListener<>() {
-                @Override
-                public void onResponse(Collection<DeleteMonitorResponse> responses) {
-                    SetOnce<RestStatus> errorStatusSupplier = new SetOnce<>();
-                    if (responses.stream().filter(response -> {
-                        if (response.getStatus() != RestStatus.OK) {
-                            log.error("Detector not being deleted because monitor [{}] could not be deleted. Status [{}]", response.getId(), response.getStatus());
-                            errorStatusSupplier.trySet(response.getStatus());
-                            return true;
+            // If detector doesn't have the workflows it means that older version of the plugin is used
```
eirsep, line 150: we can use step listener instead of repeating logic
first step is deleting workflow. if workflow doesnt exist listener can return success response or soemthing like that.
much more cleaner and understandable that way

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -253,23 +263,61 @@ private void createMonitorFromQueries(String index, List<Pair<String, Rule>> rul
         addFirstMonitorStep.whenComplete(addedFirstMonitorResponse -> {
                 monitorResponses.add(addedFirstMonitorResponse);
                 int numberOfUnprocessedResponses = monitorRequests.size() - 1;
-                if (numberOfUnprocessedResponses == 0){
-                    listener.onResponse(monitorResponses);
+
+                if (numberOfUnprocessedResponses == 0) {
+                    workflowService.upsertWorkflow(
+                        monitorResponses.stream().map(IndexMonitorResponse::getId).collect(Collectors.toList()),
+                        null,
```
eirsep, line 270: which field is being set as null?

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -253,23 +263,61 @@ private void createMonitorFromQueries(String index, List<Pair<String, Rule>> rul
...
+                                    new ActionListener<>() {
+                                        @Override
+                                        public void onResponse(IndexWorkflowResponse workflowResponse) {
+                                            detector.setWorkflowIds(List.of(workflowResponse.getId()));
+                                            listener.onResponse(monitorResponses);
+                                        }
+
+                                        @Override
+                                        public void onFailure(Exception e) {
+                                            listener.onFailure(e);
```
eirsep, line 310: delete monitors if we fail to create workflow?

eirsep, line 310: add error logs everywhere log.error("message", e)

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -369,29 +419,109 @@ private void updateAlertingMonitors(
...
+            refreshPolicy,
+            new ActionListener<>() {
+                @Override
+                public void onResponse(List<DeleteMonitorResponse> deleteMonitorResponses) {
+                    listener.onResponse(updatedMonitors);
+                }
+
+                @Override
+                public void onFailure(Exception e) {
+                    listener.onFailure(e);
```
eirsep, line 465: error logs

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -1141,7 +1230,30 @@ public void onResponse(IndexResponse response) {
 
                 @Override
                 public void onFailure(Exception e) {
-                    onFailures(e);
+                    // Revert the workflow and monitors created in previous steps
+                    workflowService.deleteWorkflow(request.getDetector().getWorkflowIds().get(0),
```
eirsep, line 1234: what if workflow never gets created

eirsep: how will we delete monitors then

src/main/java/org/opensearch/securityanalytics/util/MonitorService.java
```diff
@@ -0,0 +1,79 @@
...
+                    }
+                    return false;
+                }).count() > 0) {
+                    listener.onFailure(new OpenSearchStatusException("Monitor associated with detected could not be deleted", errorStatusSupplier.get()));
+                }
+                listener.onResponse(responses.stream().collect(Collectors.toList()));
+            }
+            @Override
+            public void onFailure(Exception e) {
+                listener.onFailure(e);
```
eirsep, line 66: error logs for `e` and `e.suppressedExceptions`

src/main/java/org/opensearch/securityanalytics/util/MonitorService.java
```diff
@@ -0,0 +1,79 @@
...
+import org.opensearch.action.support.GroupedActionListener;
+import org.opensearch.action.support.WriteRequest;
+import org.opensearch.client.Client;
+import org.opensearch.client.node.NodeClient;
+import org.opensearch.commons.alerting.AlertingPluginInterface;
+import org.opensearch.commons.alerting.action.DeleteMonitorRequest;
+import org.opensearch.commons.alerting.action.DeleteMonitorResponse;
+import org.opensearch.rest.RestStatus;
+
+public class MonitorService {
```
eirsep, line 25: java doc to explain this is an Alerting commons interface

src/main/java/org/opensearch/securityanalytics/util/WorkflowService.java
```diff
@@ -0,0 +1,159 @@
...
+import org.opensearch.commons.alerting.model.Delegate;
+import org.opensearch.commons.alerting.model.Monitor.MonitorType;
+import org.opensearch.commons.alerting.model.Sequence;
+import org.opensearch.commons.alerting.model.Workflow;
+import org.opensearch.commons.alerting.model.Workflow.WorkflowType;
+import org.opensearch.index.seqno.SequenceNumbers;
+import org.opensearch.rest.RestRequest.Method;
+import org.opensearch.securityanalytics.model.Detector;
+
+public class WorkflowService {
```
eirsep, line 37: java doc to explain this is an Alerting commons interface

src/main/resources/mappings/detectors.json
```diff
@@ -88,6 +88,9 @@
             }
           }
         },
+        "workflow_ids": {
```
eirsep, line 91: keyword or text?

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
stevanbz: we are deleting the monitors in workflowService.upsertWorkflow if upserts fails

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
stevanbz: updatedMonitors

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
stevanbz: Then it's an error in program flow - if the workflow is not created (because of the error) the detector won't be created.

stevanbz:
> what happens
> 
>     1. User creates detector in 2.6
> 
>     2. upgrades cluster to version where workflows are introduced
> 
>     3. updates detector rules? Will new workflow be created and the monitors be disabled? or will we not create workflows for existing detectors.
> 
>     4. let's closely check the udpate scenario

Check it out [here](https://github.com/opensearch-project/security-analytics/blob/13d83a2b41b213154fc8dc17252806382edcaabf/src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java#L417-L446)
You can see that we are checkign if the detector is supporting the workflow. If it does, we're [updating the workflow delegate list](https://github.com/opensearch-project/security-analytics/blob/13d83a2b41b213154fc8dc17252806382edcaabf/src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java#L470-L523) if not, we are just deleting the monitors.

src/main/java/org/opensearch/securityanalytics/model/Detector.java
stevanbz: I added list because I though it's more generic solution - which will allow us (if we consider that scenario at all in one moment) to add multiple workflows without taking care about the detectors created in the period once we had only one workflow (and creating special case branches in the code with some if-else logic). 
Does this makes sense? If not, I will remove it

src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java
```diff
@@ -137,35 +147,57 @@ public void onFailure(Exception t) {
...
+                                    public void onFailure(Exception e) {
+                                        if (counter.compareAndSet(false, true)) {
+                                            finishHim(null, e);
+                                        }
+                                    }
+                                });
                         }
-                        return false;
-                    }).count() > 0) {
-                        onFailures(new OpenSearchStatusException("Monitor associated with detected could not be deleted", errorStatusSupplier.get()));
```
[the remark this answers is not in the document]
stevanbz: It is not removed - it's extracted to monitor service. Check it out [here](https://github.com/opensearch-project/security-analytics/blob/8d2b5c95913080fda9c5f71c809edf0de86d68e0/src/main/java/org/opensearch/securityanalytics/util/MonitorService.java#L43-L79)

src/main/java/org/opensearch/securityanalytics/util/MonitorService.java
stevanbz: I thought to add something like this:
  log.error("Error deleting monitors", e);
When you say e.suppressedExceptions you mean we should also do
  log.error("Error deleting monitors",  e.getSuppressed()); or?

src/main/resources/mappings/detectors.json
stevanbz: Makes sense to be keyword

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
getsaurabh02: Can the same monitor be referred to other workflow?

eirsep:
@stevanbz 
Can we have an INTERNAL cluster setting `security_analytics.enable_workflow_usage` to turn off or turn on the feature.
If flag is enabled, we will create workflow in detector 
If flag is disabled, we will create only individual enabled monitors
then create following test:
1. disable the above setting
2. create detector (which would not have workflow)
3. enabled the above setting
4. update the workflow and verify that no workflow is created for that detector

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
stevanbz: @getsaurabh02 Yes - one monitor can be part of delegate list of two different workflows

stevanbz:
> @stevanbz Can we have an INTERNAL cluster setting `security_analytics.enable_workflow_usage` to turn off or turn on the feature. If flag is enabled, we will create workflow in detector If flag is disabled, we will create only individual enabled monitors then create following test:
> 
>     1. disable the above setting
> 
>     2. create detector (which would not have workflow)
> 
>     3. enabled the above setting
> 
>     4. update the workflow and verify that no workflow is created for that detector



> @stevanbz Can we have an INTERNAL cluster setting `security_analytics.enable_workflow_usage` to turn off or turn on the feature. If flag is enabled, we will create workflow in detector If flag is disabled, we will create only individual enabled monitors then create following test:
> 
>     1. disable the above setting
> 
>     2. create detector (which would not have workflow)
> 
>     3. enabled the above setting
> 
>     4. update the workflow and verify that no workflow is created for that detector

Sorry for the late response. Yeah sure - we can do this. Give me some time to implement and will ping you. Makes sense

stevanbz:
> @stevanbz Can we have an INTERNAL cluster setting `security_analytics.enable_workflow_usage` to turn off or turn on the feature. If flag is enabled, we will create workflow in detector If flag is disabled, we will create only individual enabled monitors then create following test:
> 
>     1. disable the above setting
> 
>     2. create detector (which would not have workflow)
> 
>     3. enabled the above setting
> 
>     4. update the workflow and verify that no workflow is created for that detector

Yeah sure. Makes sense. Will add the flag. Just one correction re the item 4.
instead of "update the workflow and verify that no workflow is created for that detector" I guess you wanted to say "update the **detector**..."

src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java
stevanbz: Done

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -546,7 +707,7 @@ private IndexMonitorRequest createBucketLevelMonitorRequest(
          triggers.add(bucketLevelTrigger1);
```
sbcd90, line 707: i still do not see the bucket level triggers passed to the workflow. will it be part of a future pr?

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
```diff
@@ -974,12 +977,472 @@ else if (ruleId == minRuleId) {
...
+        assertEquals(2, monitorIds.size());
+
+        assertNotNull("Workflow not created", detectorMap.get("workflow_ids"));
+        assertEquals("Number of workflows not correct", 1, ((List<String>) detectorMap.get("workflow_ids")).size());
+
+        // Verify workflow
+        verifyWorkflow(detectorMap, monitorIds, 2);
+    }
+
+    public void testCreateDetector_verifyWorkflowExecutionBucketLevelDocLevelMonitors_success() throws IOException {
```
sbcd90, line 1314: can this test case be create a detector with an aggregation sigma rule with a detector trigger that generates a finding but not an alert?

can we add another test case to create a detector with an aggregation sigma rule with a detector trigger that generates a finding & also an alert?

sbcd90 (changes requested):

sbcd90 (approved):

eirsep (approved):

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
eirsep: Will address this in a quick follow up PR

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
eirsep: Will address this in a quick follow up PR

PR #529, opened 2023-09-01
segment replication changes

### LINKED ISSUE #473: Compatibility with segment replication
### Summary
With 2.9.0 release, there are lot of enhancements going in for segment replication[1][2] feature (went GA in 2.7.0), we need to ensure different plugins are compatible with current state of this feature. Previously, we ran [tests](https://github.com/opensearch-project/OpenSearch/issues/6602) on plugin repos to verify this compatibility but want plugin owners to be aware of these changes so that required updates (if any) can be made. With `2.10.0` release, [remote store feature](https://github.com/opensearch-project/OpenSearch/issues/2992) is going GA which internally uses SEGMENT replication strategy only i.e. it enforces all indices to use `SEGMENT` replication strategy. So, it is important to validate plugins are compatible with segment replication feature. 

#### 1. Refresh policy behavior
1. RefreshPolicy.IMMEDIATE will only refresh primary shards but not replica shards immediately. Instead post refresh, primary will start  a round of segment replication to update the replica shard copies leading to eventual consistency.
2. RefreshPolicy.WAIT_UNTIL ensures the indexing operation is searchable in your cluster i.e. RAW (Read after write guarantee). With segment replication, this guarantee is not promised due to delay in replica shared updates from asynchronous background refreshes.

####  2. Refresh lag on replicas
With segment replication, there is inherent delay in documents to be searchable on replica shard copies. This is due to the fact that replica shard copies over data (segment) files from primary. Thus, compared to document replication, there will be on average increase in amount of time the replica shards are consistent with primaries.

#### 3. System/hidden indices support
With https://github.com/opensearch-project/OpenSearch/pull/8200, system and hidden indices are now supported with `SEGMENT` replication strategy. We need to ensure there are no bottlenecks which prevents system/hidden indices with segment replication.

### Next steps
With segment replication strong reads are not guaranteed. Thus, if the plugin needs strong reads guarantees specially as alternative to change in behavior of refresh policy and lag on replicas (point 1 and 2 above), we need to update search requests to target primary shard only. With https://github.com/opensearch-project/OpenSearch/pull/7375, core now supports primary shards only based search. Please follow [documentation](https://opensearch.org/docs/latest/api-reference/search/#the-preference-query-parameter) for examples and details

### Open questions
In case of any questions or issues, please post it in [core](https://github.com/opensearch-project/OpenSearch/issues/8211) issue

### Reference
[1] [Design](https://github.com/opensearch-project/OpenSearch/issues/2229)

[2] [Documentation](https://opensearch.org/docs/latest/tuning-your-cluster/availability-and-recovery/segment-replication/index/)

sbcd90:
### Description
[Describe what this change achieves]
 
### Issues Resolved
#473 
 
### Check List
  - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
+++ b/src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
@@ -12,6 +12,7 @@ import java.util.Optional;
 import java.util.function.Supplier;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.opensearch.cluster.routing.Preference;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.core.action.ActionResponse;
@@ -293,6 +294,7 @@ public class SecurityAnalyticsPlugin extends Plugin implements ActionPlugin, Map
         // Trigger initialization of prepackaged rules by calling SearchRule API
         SearchRequest searchRequest = new SearchRequest(Rule.PRE_PACKAGED_RULES_INDEX);
         searchRequest.source(new SearchSourceBuilder().query(QueryBuilders.matchAllQuery()).size(0));
+        searchRequest.preference(Preference.PRIMARY_FIRST.type());
         client.execute(
                 SearchRuleAction.INSTANCE,
                 new SearchRuleRequest(true, searchRequest),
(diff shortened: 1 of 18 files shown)


dreamer-89 (commented):
Thanks @sbcd90 for this quick change. 

Overall changes look good to me, please verify below before merge that for all requests where you added `_primary_first` preference:
1. Your plugin needs strong reads.
2. You are using IMMEDIATE, WAIT_UNTIL refresh policy for data ingestion. This is because, use of these refresh policies were giving strong reads guarantees before. So, if these are not used, it probably means you don't need strong reads for those search requests.

getsaurabh02 (approved):

eirsep (approved):

PR #533, opened 2023-09-05
Using alerting workflows in detectors  (#394)

eirsep:
Enables workflow creation - all the monitors related with the given detector will be part of the detector Enables workflow update during detector update
Deletes the workflow during the detector deletion
Adds additional field to a detector


sbcd90 (approved):
pr looks good. approving the pr. but we need to work on the `bucket-level trigger impl`.

getsaurabh02 (approved):

PR #539, opened 2023-09-06
Auto expand replicas

lezzago:
### Description
Auto expand replicas so for single node clusters, it wouldn't try to create a replica and cause the cluster to be yellow
 
### Issues Resolved
N/A
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/util/RuleIndices.java
+++ b/src/main/java/org/opensearch/securityanalytics/util/RuleIndices.java
@@ -5,6 +5,8 @@
 package org.opensearch.securityanalytics.util;
 
 import java.util.Set;
+
+import com.google.common.collect.ImmutableMap;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.OpenSearchStatusException;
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/securityanalytics/util/RuleIndices.java
```diff
@@ -99,9 +101,13 @@ public static String ruleMappings() throws IOException {
 
     public void initRuleIndex(ActionListener<CreateIndexResponse> actionListener, boolean isPrepackaged) throws IOException {
         if (!ruleIndexExists(isPrepackaged)) {
+            Settings indexSettings = Settings.builder()
+                    .put("index.hidden", true)
+                    .put("index.auto_expand_replicas", "0-all")
```
eirsep, line 106: was this a miss?

have we already dont this for other system indices in SAP?

lezzago: These indices are auto generated, so it would cause single node clusters starting up to immediately be in a yellow state even if they do not use the plugin.

eirsep (approved):

sbcd90 (approved):

PR #545, opened 2023-09-07
adds support for alerts and triggers on group by based sigma rules 

eirsep:
### Description
- For each group by rule a bucket level monitor is created.
- Bucket level monitor doesn't support the current detector triggers which cater focussedly to per document alerting 
- This change uses workflow's chained monitor findings feature to feed findings generated by all bucket level monitors as source data to a match all doc level monitor. The chained findings monitor simply creates findings for all docs which match the bucket level monitor rule and honors the detector trigger condition creating alerts if there are any group by rules mentioned in the trigger.

Behaviour of non-group by rule matching is unaffected.

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
+++ b/src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
@@ -359,7 +359,7 @@ public class TransportIndexDetectorAction extends HandledTransportAction<IndexDe
     ) {
         if (enabledWorkflowUsage) {
             workflowService.upsertWorkflow(
-                monitorResponses.stream().map(IndexMonitorResponse::getId).collect(Collectors.toList()),
+                monitorResponses,
                 null,
                 detector,
                 refreshPolicy,
@@ -596,8 +596,8 @@ public class TransportIndexDetectorAction extends HandledTransportAction<IndexDe
         } else {
             // Update workflow and delete the monitors
             workflowService.upsertWorkflow(
-                addedMonitorIds,
-                updatedMonitorIds,
+                addNewMonitorsResponse,
+                updateMonitorResponse,
                 detector,
                 refreshPolicy,
                 detector.getWorkflowIds().get(0),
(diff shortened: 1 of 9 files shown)


src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
```diff
@@ -940,7 +941,7 @@ else if (ruleId == minRuleId) {
         // Assert findings
         assertNotNull(getFindingsBody);
         // 8 findings from doc level rules, and 3 findings for aggregation (sum, max and min)
-        assertEquals(11, getFindingsBody.get("total_findings"));
+        assertEquals(19, getFindingsBody.get("total_findings"));
```
sbcd90, line 944: should we hide the findings generated by match of `_id: *` queries?

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
```diff
@@ -1409,7 +1410,7 @@ public void testCreateDetector_verifyWorkflowExecutionBucketLevelDocLevelMonitor
         Map<String, Object> getFindingsBody = entityAsMap(getFindingsResponse);
 
         assertNotNull(getFindingsBody);
-        assertEquals(6, getFindingsBody.get("total_findings"));
+        assertEquals(10, getFindingsBody.get("total_findings"));
```
sbcd90, line 1413: same comment as https://github.com/opensearch-project/security-analytics/pull/545/files#r1319008403

eirsep: let's go with this for now
will do de-dupe later.

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
eirsep: been thinking about this

let's go with this for now
will do de-dupe later.

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -785,7 +850,7 @@ private IndexMonitorRequest createBucketLevelMonitorRequest(
          triggers.add(bucketLevelTrigger1);
          } **/
 
-        Monitor monitor = new Monitor(monitorId, Monitor.NO_VERSION, detector.getName(), false, detector.getSchedule(), detector.getLastUpdateTime(), null,
+        Monitor monitor = new Monitor(monitorId, Monitor.NO_VERSION, detector.getName() + UUID.randomUUID(), false, detector.getSchedule(), detector.getLastUpdateTime(), null,
```
sbcd90, line 853: is this change needed? i reverted this & integ tests still work fine.

sbcd90 (approved):

eirsep:
https://github.com/opensearch-project/security-analytics/issues/558

getsaurabh02 (approved):

PR #669, opened 2023-10-16
Integrate threat intel feeds

eirsep:
### Description
---
 **Job Scheduler and threat intel feed Integration**
--
 - Integrates Job Scheduler Plugin in Security Analytics
 - Adds framework in Security Analytics for configuring pre-package feeds.
 - Adds support for job to constantly update system indices with feed data
 - Configures 1 free feed - otx alienvault ip reputation ( a feed of malicious IPs)
 - Adds support for IoC(indicator of Compromise) To ECS Field mapping for each log type which tells us fields where the given type of IoC may occur
  - Feed data is stored in system indices. new data is populated in new system index and the old index is deleted.
  - ---
  **Detector creation/updation flow changes**
--
 - Adds framework to convert threat intel IoCs data into doc level queries
 -  new boolean field added to Detector model - `threat_intel_enabled`. (Value is False by default). If set to true, we will create threat intel data based queries
 - When feed data is updated, all detectors having threat intel data are also updated with queries built from the fresh feeds
 - Doc level monitor created to execute sigma rules will have new queries.
 ----
**Detector trigger**:
--
- Detector trigger now supports new field `detection_types` which is a list with 2 possible values "rules" and "threat_intel"
- If trigger is configured and detection_types is passed with only "rules" in the list , we will create alerts when documents match sigma rules.
- If detection_types is passed with only "threat_intel" in the list , we will create alerts when documents match threat intel based queries.
- If detection_types is passed with both "rules" and "threat_intel" in the list , we will create alerts when documents  match either sigma rules or threat intel based queries
- Detection types list param cannot be empty.
----
**Detector findings and threat_intel_based queries**:
--
- Findings created from matching threat intel based queries will contain the tags `threat_intel`, `ioc_type:<ioc_type>`, "`field:<field>`, `feed_name:<feed_name>`
- Threat intel based queries will have query id starting with the prefix `threat_intel`
- Threat intel based queries will have severity level `high`

----
**Get Mapping VIew changes**:
--
- We have added a new field `threat_intel_field_aliases` which returns a list of field aliases per ioc for each log type that the user needs to provide mappings for, from his source data.
  
### Issues Resolved
[Threat intel feeds in Security Analytics](https://github.com/opensearch-project/security-analytics/issues/672)


build.gradle
```diff
@@ -69,6 +69,7 @@ opensearchplugin {
     name 'opensearch-security-analytics'
     description 'OpenSearch Security Analytics plugin'
     classname 'org.opensearch.securityanalytics.SecurityAnalyticsPlugin'
+//    extendedPlugins = ['opensearch-job-scheduler']
```
lezzago, line 72: dev code

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
```diff
@@ -4,20 +4,13 @@
  */
 package org.opensearch.securityanalytics;
 
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
lezzago, line 7: bad to have * imports

src/main/java/org/opensearch/securityanalytics/model/ThreatIntelFeedData.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    private XContentBuilder createXContentBuilder(XContentBuilder builder, ToXContent.Params params) throws IOException {
+        builder.startObject();
+        if (params.paramAsBoolean("with_type", false)) {
+            builder.startObject(type);
+        }
+        builder.field(TYPE_FIELD, type);
+        builder
+                .field(IOC_TYPE_FIELD, iocType)
```
lezzago, line 143: nitpick: Why not use the same builder and add more fields like it has been done with the second builder.

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
```diff
@@ -0,0 +1,133 @@
...
+        for (Map.Entry<String, Set<String>> entry : iocTypeToValues.entrySet()) {
+            String query = buildQueryStringQueryWithIocList(iocs);
+            List<String> fields = iocFieldList.stream().filter(t -> entry.getKey().matches(t.getIoc())).findFirst().get().getFields();
+
+            // create doc
+            for (String field : fields) { //todo increase max clause count from 1024
+                queries.add(new DocLevelQuery(
+                        constructId(detector, entry.getKey()), tifdList.get(0).getFeedId(),
+                        Collections.emptyList(),
+                        "windows-hostname:(120.85.114.146 OR 103.104.106.223 OR 185.191.246.45 OR 120.86.237.94)",
```
lezzago, line 65: lets store these important IPs in a const

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
```diff
@@ -0,0 +1,295 @@
...
+            }
+            //if index exists
+            String tifdIndex = IndexUtils.getNewIndexByCreationDate(
+                    this.clusterService.state(),
+                    this.indexNameExpressionResolver,
+                    ".opensearch-sap-threatintel*"
+            );
+
+            SearchRequest searchRequest = new SearchRequest(tifdIndex);
+            searchRequest.source().size(9999); //TODO: convert to scroll
```
lezzago, line 121: is this going to be done in a follow up? If so, please create an issue and link it here.

src/main/java/org/opensearch/securityanalytics/threatIntel/feedMetadata/TIFMetadataService.java
lezzago, line 1: This is an empty file

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFJobParameterService.java
```diff
@@ -0,0 +1,198 @@
...
+    private String getIndexMapping() {
+        try {
+            try (InputStream is = TIFJobParameterService.class.getResourceAsStream("/mappings/threat_intel_job_mapping.json")) {
+                try (BufferedReader reader = new BufferedReader(new InputStreamReader(is, StandardCharsets.UTF_8))) {
+                    return reader.lines().map(String::trim).collect(Collectors.joining());
+                }
+            }
+        } catch (IOException e) {
+            log.error("Runtime exception", e);
+            throw new SecurityAnalyticsException("Runtime exception", RestStatus.INTERNAL_SERVER_ERROR, e); //TODO
```
lezzago, line 100: there is a TODO here

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFJobParameterService.java
```diff
@@ -0,0 +1,198 @@
...
+            try {
+                return client.prepareIndex(TIFJobExtension.JOB_INDEX_NAME)
+                        .setId(jobSchedulerParameter.getName())
+                        .setOpType(DocWriteRequest.OpType.INDEX)
+                        .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE)
+                        .setSource(jobSchedulerParameter.toXContent(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS))
+                        .execute()
+                        .actionGet(clusterSettings.get(SecurityAnalyticsSettings.THREAT_INTEL_TIMEOUT));
+            } catch (IOException e) {
+                throw new SecurityAnalyticsException("Runtime exception", RestStatus.INTERNAL_SERVER_ERROR, e); //TODO
```
lezzago, line 121: TODO left here

src/test/java/org/opensearch/securityanalytics/threatIntel/integTests/ThreatIntelJobRunnerIT.java
```diff
@@ -0,0 +1,154 @@
...
+    private List<String> getThreatIntelFeedIds(int num) throws IOException {
+        String request = getMatchAllSearchRequestString(num);
+        SearchResponse res = executeSearchAndGetResponse(".opensearch-sap-threatintel*", request, false);
+        return getTifdList(res, xContentRegistry()).stream().map(it -> it.getFeedId()).collect(Collectors.toList());
+    }
+
+//    private String getJobSchedulerDoc(int num) throws IOException {
+//        String request = getMatchAllSearchRequestString(num);
+//        SearchResponse res = executeSearchAndGetResponse(".scheduler-sap-threatintel-job*", request, false);
+//    }
```
lezzago, line 142: Should we remove commented code?

build.gradle
eirsep: @jowg-amazon

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
eirsep: @jowg-amazon

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
eirsep: fixed this. this was a miss.

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFJobExtension.java
```diff
@@ -0,0 +1,46 @@
...
+import org.opensearch.jobscheduler.spi.ScheduledJobParser;
+import org.opensearch.jobscheduler.spi.ScheduledJobRunner;
+
+import java.util.Map;
+
+public class TIFJobExtension implements org.opensearch.jobscheduler.spi.JobSchedulerExtension {
+    /**
+     * Job index name for a TIF job
+     */
+    public static final String JOB_INDEX_NAME = ".scheduler-sap-threatintel-job";
```
eirsep, line 17: .opensearch-sap-threatintel-job

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFJobExtension.java
```diff
@@ -0,0 +1,46 @@
...
+     * Job index setting
+     *
+     * We want it to be single shard so that job can be run only in a single node by job scheduler.
+     * We want it to expand to all replicas so that querying to this index can be done locally to reduce latency.
+     */
+    public static final Map<String, Object> INDEX_SETTING = Map.of("index.number_of_shards", 1, "index.auto_expand_replicas", "0-all", "index.hidden", true);
+
+    @Override
+    public String getJobType() {
+        return "scheduler_sap_threatintel_job";
```
eirsep, line 29: remove "scheduler" prefix or replace with opensearch.

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
eirsep: @jowg-amazon

src/test/java/org/opensearch/securityanalytics/threatIntel/integTests/ThreatIntelJobRunnerIT.java
eirsep: this test will be fixed

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFJobParameterService.java
eirsep: @jowg-amazon

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFJobParameterService.java
eirsep: @jowg-amazon

src/main/java/org/opensearch/securityanalytics/threatIntel/feedMetadata/TIFMetadataService.java
eirsep: removed

src/main/java/org/opensearch/securityanalytics/model/ThreatIntelFeedData.java
eirsep: i didn't understand the comment. there is only one builder object

lezzago: Sorry, I meant more for formatting so making this like:
```
builder.field(TYPE_FIELD, type)
                .field(IOC_TYPE_FIELD, iocType)
                .field(IOC_VALUE_FIELD, iocValue)
                .field(FEED_ID_FIELD, feedId)
                .timeField(TIMESTAMP_FIELD, String.format(Locale.getDefault(), "%s_in_millis", TIMESTAMP_FIELD), timestamp.toEpochMilli());
```

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
```diff
@@ -0,0 +1,172 @@
...
+                        ThreatIntelFeedData::getIocType,
+                        Collectors.mapping(ThreatIntelFeedData::getIocValue, Collectors.toSet())
+                ));
+
+        for (Map.Entry<String, Set<String>> entry : iocTypeToValues.entrySet()) {
+            String query = buildQueryStringQueryWithIocList(iocs);
+            List<String> fields = iocFieldList.stream().filter(t -> entry.getKey().matches(t.getIoc())).findFirst().get().getFields();
+
+            // create doc
+            for (String field : fields) { //todo increase max clause count from 1024
```
lezzago, line 77: Is this TODO needed to be fixed now? If not, can we have a tracking issue?

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
```diff
@@ -0,0 +1,172 @@
...
+            sb.append(ioc);
+
+        }
+        sb.append(")");
+        return sb.toString();
+    }
+
+    public void createDocLevelQueryFromThreatIntel(List<LogType.IocFields> iocFieldList, Detector detector, ActionListener<List<DocLevelQuery>> listener) {
+        try {
+            if (false == detector.getThreatIntelEnabled() || iocFieldList.isEmpty()) {
```
lezzago, line 107: nitpick: `false == detector.getThreatIntelEnabled()` should be `!detector.getThreatIntelEnabled()`

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
```diff
@@ -0,0 +1,172 @@
...
+
+    }
+
+    private static String constructId(Detector detector, String iocType) {
+        return detector.getName() + "_threat_intel_" + iocType + "_" + UUID.randomUUID();
+    }
+
+    public void updateDetectorsWithLatestThreatIntelRules() {
+        //todo : fix query for fetching detectors with threat intel enabled = true
+//     String searchReq = "{ \"query\": { \"match\": { \"detector.threat_intel_enabled\": true } } }";
```
lezzago, line 148: Needs to be fixed?

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
```diff
@@ -0,0 +1,172 @@
...
+        SearchRequest searchRequest = new SearchRequest(DETECTORS_INDEX);
+        SearchSourceBuilder ssb = searchRequest.source();
+        ssb.size(9999);
```
lezzago, line 151: Do we need to paginate?

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
```diff
@@ -0,0 +1,297 @@
...
+            indexRequest.opType(DocWriteRequest.OpType.INDEX);
+            bulkRequest.add(indexRequest);
+
+            if (bulkRequest.requests().size() == batchSize) {
+                saveTifds(bulkRequest, timeout);
+            }
+        }
+        saveTifds(bulkRequest, timeout);
+        renewLock.run();
+        freezeIndex(indexName);
```
lezzago, line 224: Why do we need to freeze the index?

src/main/java/org/opensearch/securityanalytics/threatIntel/action/TransportDeleteTIFJobAction.java
lezzago, line 1: Why does this need to be a Transport action? This should be a utility since this is only for SAP to use.

src/main/java/org/opensearch/securityanalytics/threatIntel/action/RestPutTIFJobHandler.java
```diff
@@ -0,0 +1,65 @@
...
+ *     "id": {id},
+ *     "name": {name},
+ *     "update_interval_in_days": 1
+ * }
+ *
+ * When request is received, it will create a TIFjob
+ * After the creation of TIFjob is completed, it will schedule the next update task after update_interval_in_days.
+ *
+ */
+public class RestPutTIFJobHandler extends BaseRestHandler {
```
lezzago, line 39: Why do we want an API for users to use this REST API, was it just for testing purposes?

src/main/java/org/opensearch/securityanalytics/threatIntel/action/TransportPutTIFJobAction.java
lezzago, line 1: Why does this need to be a Transport action? This should be a utility since this is only for SAP to use.

build.gradle
```diff
@@ -158,17 +153,14 @@ dependencies {
     api "org.opensearch:common-utils:${common_utils_version}@jar"
     api "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
     implementation "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
+    compileOnly "org.opensearch:opensearch-job-scheduler-spi:${opensearch_build}"
+    implementation "org.apache.commons:commons-csv:1.10.0"
```
sandeshkr419, line 157: nit: Can we utilize a version from variables above instead of hardcoding it a constant 1.10.0

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
```diff
@@ -91,6 +97,8 @@ public class SecurityAnalyticsPlugin extends Plugin implements ActionPlugin, Map
     public static final String CORRELATION_RULES_BASE_URI = PLUGINS_BASE_URI + "/correlation/rules";
 
     public static final String CUSTOM_LOG_TYPE_URI = PLUGINS_BASE_URI + "/logtype";
+    public static final String JOB_INDEX_NAME = ".opensearch-sap-threatintel-job";
```
sandeshkr419, line 100: threat and intel seems to be different words? Should we use `".opensearch-sap-threat-intel-job"` instead?

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
```diff
@@ -113,8 +121,12 @@ public class SecurityAnalyticsPlugin extends Plugin implements ActionPlugin, Map
     private BuiltinLogTypeLoader builtinLogTypeLoader;
 
     private LogTypeService logTypeService;
+    @Override
+    public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings settings){
+        return List.of(new SystemIndexDescriptor(THREAT_INTEL_DATA_INDEX_NAME_PREFIX, "System index used for threat intel data"));
```
sandeshkr419, line 126: nit: use Singleton list or immutable list.

src/main/java/org/opensearch/securityanalytics/config/monitors/opensearch_security.policy
```diff
@@ -0,0 +1,11 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+grant {
+    permission java.lang.management.ManagementPermission "reputation.alienvault.com:443" "connect,resolve";
```
sandeshkr419, line 10: Doubt: Why do you need this?

src/main/java/org/opensearch/securityanalytics/model/LogType.java
```diff
@@ -62,6 +67,17 @@ public LogType(Map<String, Object> logTypeAsMap) {
...
             ).collect(Collectors.toList());
         }
+        if(logTypeAsMap.containsKey(IOC_FIELDS)) {
+            List<Map<String, Object>> iocFieldsList = (List<Map<String, Object>>) logTypeAsMap.get(IOC_FIELDS);
+            if (iocFieldsList.size() > 0) {
+                this.iocFieldsList = new ArrayList<>(mappings.size());
+                this.iocFieldsList = iocFieldsList.stream().map(e ->
+                        new IocFields(e.get(IOC).toString(), (List<String>) e.get(FIELDS))
+                ).collect(Collectors.toList());
+            }
```
sandeshkr419, line 77: Simplify:



```
            if (iocFieldsList.size() > 0) {
                this.iocFieldsList = new ArrayList<>(mappings.size());
                this.iocFieldsList = iocFieldsList.stream().map(e ->
                        new IocFields(e.get(IOC).toString(), (List<String>) e.get(FIELDS))
                ).collect(Collectors.toList());
            }
```

to

```
this.iocFieldsList = iocFieldsList.stream()
    .map(e -> new IocFields(e.get(IOC).toString(), (List<String>) e.get(FIELDS)))
    .collect(Collectors.toList());
```

to improve readbaility and avoid unnecessary checks.

src/main/java/org/opensearch/securityanalytics/model/LogType.java
```diff
@@ -134,4 +155,43 @@ public static Mapping readFrom(StreamInput sin) throws IOException {
         }
     }
 
+    /**
+     * stores information of list of field names that contain information for given IoC (Indicator of Compromise).
+     */
+    public static class IocFields implements Writeable {
+        private final String ioc;
```
sandeshkr419, line 162: nitpick: new line inconsistency in this file looks weird. No new line after class, multiple new lines between 1-2 methods, 2 new lines at end of class.

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import static org.opensearch.securityanalytics.model.Detector.DETECTORS_INDEX;
+import static org.opensearch.securityanalytics.util.DetectorUtils.getDetectors;
+
+
+public class DetectorThreatIntelService {
```
sandeshkr419, line 40: Can we add brief javadocs for classes to explain their role?

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
sandeshkr419: @lezzago `false == detector.getThreatIntelEnabled() ` is consistent wiht openSearch usage - let's keep that as it is.

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    public void updateDetectorsWithLatestThreatIntelRules() {
+        //todo : fix query for fetching detectors with threat intel enabled = true
+//     String searchReq = "{ \"query\": { \"match\": { \"detector.threat_intel_enabled\": true } } }";
+        SearchRequest searchRequest = new SearchRequest(DETECTORS_INDEX);
+        SearchSourceBuilder ssb = searchRequest.source();
+        ssb.size(9999);
+        client.execute(SearchDetectorAction.INSTANCE, new SearchDetectorRequest(new SearchRequest().source(ssb)),
+                ActionListener.wrap(r -> {
```
sandeshkr419, line 153: Can we replace `r` with something meanigful?

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
```diff
@@ -0,0 +1,297 @@
...
+ * Service to handle CRUD operations on Threat Intel Feed Data
+ */
+public class ThreatIntelFeedDataService {
+    private static final Logger log = LogManager.getLogger(ThreatIntelFeedDataService.class);
+
+    private final Client client;
+    private final IndexNameExpressionResolver indexNameExpressionResolver;
+
+    private static final Map<String, Object> INDEX_SETTING_TO_CREATE = Map.of(
+            "index.number_of_shards",
```
sandeshkr419, line 71: Don't we have some constants already in OpenSearch for `"index.number_of_shards"` and other settings? Can we utilize those constants instead of creating new strings.
`

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
```diff
@@ -0,0 +1,297 @@
...
+            "index.hidden",
+            true
+    );
+    private static final Map<String, Object> INDEX_SETTING_TO_FREEZE = Map.of(
+            "index.auto_expand_replicas",
+            "0-all",
+            "index.blocks.write",
+            true
+    );
+    private final ClusterService clusterService;
```
sandeshkr419, line 86: Can we move all variables/constants in the top of class, followed by high level methods followed by private methods towards the last? Makes it easy to visualize the class.

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    private final NamedXContentRegistry xContentRegistry;
+
+    public void getThreatIntelFeedData(
+            ActionListener<List<ThreatIntelFeedData>> listener
+    ) {
+        try {
+            //if index not exists
+            if (IndexUtils.getNewIndexByCreationDate(
```
sandeshkr419, line 108: Not sure why do we need to invoke `getNewIndexByCreationDate` if we only need to check for existence of an index. Aren't there any faster APIs/utility available?

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
```diff
@@ -0,0 +1,297 @@
...
+                    this.clusterService.state(),
+                    this.indexNameExpressionResolver,
+                    ".opensearch-sap-threatintel*"
+            ) == null) {
+                createThreatIntelFeedData();
+            }
+            //if index exists
+            String tifdIndex = IndexUtils.getNewIndexByCreationDate(
```
sandeshkr419, line 116: This si the same call as above - can we avoid making the same call twice and refactor to make only one call.

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
```diff
@@ -0,0 +1,297 @@
...
+
+            SearchRequest searchRequest = new SearchRequest(tifdIndex);
+            searchRequest.source().size(9999); //TODO: convert to scroll
+            client.search(searchRequest, ActionListener.wrap(r -> listener.onResponse(ThreatIntelFeedDataUtils.getTifdList(r, xContentRegistry)), e -> {
+                log.error(String.format(
+                        "Failed to fetch threat intel feed data from system index %s", tifdIndex), e);
+                listener.onFailure(e);
+            }));
+        } catch (InterruptedException e) {
+            log.error("failed to get threat intel feed data", e);
```
sandeshkr419, line 130: nit- remove inconsistency in using upper/lower case in error messages. eg -`Failed` in lines above, `failed` here.

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
```diff
@@ -0,0 +1,297 @@
...
+        } catch (OpenSearchException e) {
+            log.error("failed to save threat intel feed data", e);
+        }
+
+    }
+
+    private void freezeIndex(final String indexName) {
+        TimeValue timeout = clusterSettings.get(SecurityAnalyticsSettings.THREAT_INTEL_TIMEOUT);
+        StashedThreadContext.run(client, () -> {
+            client.admin().indices().prepareForceMerge(indexName).setMaxNumSegments(1).execute().actionGet(timeout);
```
sandeshkr419, line 256: Can we avoid 3 calls to ` client.admin().indices().` by storing the indices in 1 call if this doesn't affects functionality.

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
```diff
@@ -0,0 +1,172 @@
...
+                                client.execute(IndexDetectorAction.INSTANCE, new IndexDetectorRequest(
+                                                detector.getId(), WriteRequest.RefreshPolicy.IMMEDIATE,
+                                                RestRequest.Method.PUT,
+                                                detector),
+                                        ActionListener.wrap(
+                                                res -> log.debug("updated {} with latest threat intel info", res.getDetector().getId()),
+                                                e -> log.error(() -> new ParameterizedMessage("Failed to update detector {} with latest threat intel info", detector.getId()), e)));
+                            }
+                    );
+                }, e -> {
```
lezzago, line 166: should we do a bulk put call?

src/main/java/org/opensearch/securityanalytics/threatIntel/action/PutTIFJobRequest.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
+        ActionRequestValidationException errors = new ActionRequestValidationException();
+        List<String> errorMsgs = VALIDATOR.validateTIFJobName(name);
+        if (errorMsgs.isEmpty() == false) {
+            errorMsgs.stream().forEach(msg -> errors.addValidationError(msg));
+        }
+        return errors.validationErrors().isEmpty() ? null : errors;
```
sandeshkr419, line 82: Why not return only validation errors instead of all?

src/main/java/org/opensearch/securityanalytics/threatIntel/action/RestPutTIFJobHandler.java
```diff
@@ -0,0 +1,65 @@
...
+    protected RestChannelConsumer prepareRequest(final RestRequest request, final NodeClient client) throws IOException {
+        final PutTIFJobRequest putTIFJobRequest = new PutTIFJobRequest("jobname",
+                new TimeValue(1, TimeUnit.MINUTES));
+
+        return channel -> client.executeLocally(PutTIFJobAction.INSTANCE, putTIFJobRequest, new RestToXContentListener<>(channel));
+    }
+
+    @Override
+    public List<Route> routes() {
+        String path = "/_p/_s";
```
sandeshkr419, line 62: Just "_p/_s" - something seems incomplete?

src/main/java/org/opensearch/securityanalytics/threatIntel/common/ParameterValidator.java
```diff
@@ -0,0 +1,58 @@
...
+
+        if (!Strings.validFileName(tifJobName)) {
+            errorMsgs.add(
+                    String.format(Locale.ROOT, "threat intel feed job name must not contain the following characters %s", Strings.INVALID_FILENAME_CHARS)
+            );
+        }
+        if (tifJobName.contains("#")) {
+            errorMsgs.add("threat intel feed job name must not contain '#'");
+        }
+        if (tifJobName.contains(":")) {
```
sandeshkr419, line 43: Can combine `#` and `:` conditions?

src/main/java/org/opensearch/securityanalytics/threatIntel/common/TIFMetadata.java
```diff
@@ -0,0 +1,225 @@
...
+        out.writeString(name);
+        out.writeString(organization);
+        out.writeString(description);
+        out.writeString(feedType);
+        out.writeString(iocType);
+        out.writeInt(iocCol);
+        out.writeBoolean(hasHeader);
+    }
+
+    private TIFMetadata() {
```
sandeshkr419, line 205: Why do we need this?

src/main/java/org/opensearch/securityanalytics/threatIntel/feedMetadata/BuiltInTIFMetadataLoader.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    protected void doStart() {
+        ensureTifMetadataLoaded();
+    }
+
+    @Override
+    protected void doStop() {
+
```
sandeshkr419, line 111: no actions?

src/main/java/org/opensearch/securityanalytics/threatIntel/feedMetadata/BuiltInTIFMetadataLoader.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    protected void doClose() throws IOException {
+
```
sandeshkr419, line 116: no actions?

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFJobParameter.java
```diff
@@ -0,0 +1,557 @@
...
+        this.state = state;
+    }
+
+    /**
+     * tif job parser
+     */
+    public static final ConstructingObjectParser<TIFJobParameter, Void> PARSER = new ConstructingObjectParser<>(
+            "tifjob_metadata",
+            true,
+            args -> {
```
sandeshkr419, line 189: not checking args length?

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
eirsep: update detector entails updating monitor's doc level queries
hence i cant add a bulk call here before alerting supports bulk PUT

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
```diff
@@ -0,0 +1,302 @@
...
+        );
+
+        if (response.isAcknowledged() == false) {
+            throw new OpenSearchException("failed to delete data[{}]", String.join(",", indices));
+        }
+    }
+
+    private void createThreatIntelFeedData() throws InterruptedException {
+        CountDownLatch countDownLatch = new CountDownLatch(1);
+        client.execute(PutTIFJobAction.INSTANCE, new PutTIFJobRequest("feed_updater", clusterSettings.get(SecurityAnalyticsSettings.TIF_UPDATE_INTERVAL))).actionGet();
```
bowenlan-amzn, line 267: Seems only one job is needed for download threat intel? 
If so, don't use job scheduler, just do a sheduled job using ThreadPool like 
https://github.com/opensearch-project/OpenSearch/blob/41a12e28fff58c7262bfeb725c8a31e5ffa266f3/server/src/main/java/org/opensearch/threadpool/ThreadPool.java#L485-L496

job scheduler support lots of jobs of same type, if for every detector or maybe every threat intel source, you want to create job, then it would make more sense

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
```diff
@@ -38,18 +36,15 @@
...
-import org.opensearch.plugins.ClusterPlugin;
-import org.opensearch.plugins.EnginePlugin;
-import org.opensearch.plugins.MapperPlugin;
-import org.opensearch.plugins.Plugin;
-import org.opensearch.plugins.SearchPlugin;
+import org.opensearch.indices.SystemIndexDescriptor;
+import org.opensearch.jobscheduler.spi.JobSchedulerExtension;
+import org.opensearch.jobscheduler.spi.ScheduledJobParser;
+import org.opensearch.jobscheduler.spi.ScheduledJobRunner;
+import org.opensearch.plugins.*;
```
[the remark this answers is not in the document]
sandeshkr419: I think this package has multiple `*` imports. We should raise a separate PR to fix all instances. But we can fix this import here as well.

build.gradle
eirsep: we are introducing the dependency. we are not deriving from Opensearch core or any other dependency

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
eirsep: no need of todo for now

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
eirsep: `false ==` is more readable because you mentally map the condition following it as something not happening while `!` is obscure when browsing through code

src/main/java/org/opensearch/securityanalytics/threatIntel/action/TransportPutTIFJobAction.java
eirsep: Will add todo to remove all the transport action code

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
eirsep: threat intel feed updation job creates an entirely new index when it fetches latest threat intel data and deletes the old index.
if the feed data is being fetched by a detector to create queries during the above process when new index is created and old hasn't been deleted we want to ensure we pick data from the latest available index only.

eirsep: this method also helps us check if index exists at all.. as we pass in an a prefix only

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
eirsep: we have removed the method itself

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
eirsep: refactored this a bit

 but first time we fetch latest index based on regex (threat intel system index prefix) and if it doesn't exist we populate threat intel data and then we re-fetch the index

src/main/java/org/opensearch/securityanalytics/threatIntel/action/RestPutTIFJobHandler.java
eirsep: this class has been removed. it was for testing purpose

src/main/java/org/opensearch/securityanalytics/threatIntel/feedMetadata/BuiltInTIFMetadataLoader.java
eirsep: no actions here. it's interface contract

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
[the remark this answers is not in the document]
eirsep: i have fixed wildcard imports that we have introduced in this PR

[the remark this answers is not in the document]
eirsep: let's do a sanity sweep in a follow up PR across code base

src/main/java/org/opensearch/securityanalytics/threatIntel/action/TransportPutTIFJobAction.java
```diff
@@ -0,0 +1,180 @@
...
+    protected void internalDoExecute(
+            final PutTIFJobRequest request,
+            final LockModel lock,
+            final ActionListener<AcknowledgedResponse> listener
+    ) {
+        StepListener<Void> createIndexStep = new StepListener<>();
+        tifJobParameterService.createJobIndexIfNotExists(createIndexStep);
+        createIndexStep.whenComplete(v -> {
+            TIFJobParameter tifJobParameter = TIFJobParameter.Builder.build(request);
+            tifJobParameterService.saveTIFJobParameter(tifJobParameter, postIndexingTifJobParameter(tifJobParameter, lock, listener));
```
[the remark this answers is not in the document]
eirsep: +1
added todo to refactor
will follow up this pr with refactor pr

src/main/java/org/opensearch/securityanalytics/threatIntel/action/TransportPutTIFJobAction.java
```diff
@@ -0,0 +1,180 @@
...
+    ) {
+        return new ActionListener<>() {
+            @Override
+            public void onResponse(final IndexResponse indexResponse) {
+                // This is user initiated request. Therefore, we want to handle the first tifJobParameter update task in a generic thread
+                // pool.
+                threadPool.generic().submit(() -> {
+                    AtomicReference<LockModel> lockReference = new AtomicReference<>(lock);
+                    try {
+                        createThreatIntelFeedData(tifJobParameter, lockService.getRenewLockRunnable(lockReference));
```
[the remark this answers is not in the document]
eirsep: +1
added todo to refactor
will follow up this pr with refactor pr

src/main/java/org/opensearch/securityanalytics/threatIntel/ThreatIntelFeedDataService.java
eirsep: i disagree due to following reasons:
1. we should use job scheduler because it provides out of box job scheduling at cluster level already. detectors are created on data nodes and in a distributed system with mutliple data nodes it is not a safe approach for every single data node to simply run this job
2. when we add more feeds and custom feeds job scheduler is the only component that can help us schedule/deschedule multiple jobs. it gives flexibility of using mutliple frequencies. 
3. Jobs are stored in system index and hence will persist through upgrades and are snapshotted.
4. only one node will run the job and this scheduling and delegation is provided out of box. we should not re-invent job delegation between data nodes when we have a centralized solution
5. Job scheduler would be used for sigma rule updation in future too which is a big customer ask right now

Dockerfile
lezzago, line 1: remove

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
```diff
@@ -0,0 +1,203 @@
...
+                ));
+
+        for (Map.Entry<String, Set<String>> entry : iocTypeToValues.entrySet()) {
+            String query = buildQueryStringQueryWithIocList(iocs);
+            List<String> fields = iocFieldList.stream().filter(t -> entry.getKey().matches(t.getIoc())).findFirst().get().getFields();
+
+            // create doc
+            for (String field : fields) {
+                queries.add(new DocLevelQuery(
+                        constructId(detector, entry.getKey()), tifdList.get(0).getFeedId(),
```
lezzago, line 84: Why are we getting the feed id based on the first item in `tifdList`? Is it the same throughout?

praveensameneni (approved):

lezzago (approved):

src/main/java/org/opensearch/securityanalytics/threatIntel/DetectorThreatIntelService.java
eirsep: Currently yes,

PR #725, opened 2023-11-22
#683 Sigma keywords field not handled correctly

### LINKED ISSUE #683: [BUG] Sigma keywords field not handled correctly
**What is the bug?**
Sigma rules with "keywords" as a field under detection section generate broken queries where the field being queried is an index number (0, 1, 2 and so on) 

Example:
Possible CVE-2021-1675 Print Spooler Exploitation (A prepackaged rule) has this Sigma Detection field:
image

  The query is then translated to this:
  `((((winlog.event_id: 808) OR (winlog.event_id: 4909)) AND ((winlog.event_data.ErrorCode: "0x45A") OR (winlog.event_data.ErrorCode: "0x7e"))) OR ((_21: "The_ws_print_ws_spooler_ws_failed_ws_to_ws_load_ws_a_ws_plug\-in_ws_module") OR (_22: "MyExploit.dll") OR (_23: "evil.dll") OR (_24: "\\addCube.dll") OR (_25: "\\rev.dll") OR (_26: "\\rev2.dll") OR (_27: "\\main64.dll") OR (_28: "\\mimilib.dll") OR (_29: "\\mimispool.dll"))) AND ((NOT _30: "_ws_registration_ws_timed_ws_out"))`
 Where the fields from keyword are incremental numbers preceded by _  (This bug happens for multiple rules, I am pulling just one out of sequence here)
 
 The same bug happens when you duplicate this standard rule with slightly different results.
 The YAML for the rule after duplication is changed to this:
 
image

Which is reflected in GUI like this:
image

 This also occurs when you upload a rule via the custom rule API.

1. Go to security analytics > detectors > detection rules
2. Find CVE-2021-1675 Print Spooler Exploitation
3. Duplicate Rule
4. View error in both GUI and YAML Editor. You can also view the created broken query by searching for the rule:
POST /_plugins/_security_analytics/rules/_search?pre_packaged=true

Sigma rules with the keywords field under detection section to be properly translated to a query that can be used by the Security Analytics plugin.

 - OS 2.11 fresh install running security analytics plugin

Thanks,
Jay

jowg-amazon:
### Description
Changes the query construction of keywords in sigma rules from ```(_0:"value")``` to ```("value")``` to create a query string query without a designated field so that the keywords are instead matched against all the fields in a document.

### Issues Resolved
#683 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/test/java/org/opensearch/securityanalytics/TestHelpers.java
+++ b/src/test/java/org/opensearch/securityanalytics/TestHelpers.java
@@ -379,6 +379,68 @@ public class TestHelpers {
                 "level: high";
     }
 
+    public static String randomRuleWithStringKeywords() {
+        return "title: Remote Encrypting File System Abuse\n" +
+                "id: 5f92fff9-82e2-48eb-8fc1-8b133556a551\n" +
+                "description: Detects remote RPC calls to possibly abuse remote encryption service via MS-EFSR\n" +
+                "references:\n" +
+                "    - https://attack.mitre.org/tactics/TA0008/\n" +
+                "    - https://msrc.microsoft.com/update-guide/vulnerability/CVE-2021-36942\n" +
+                "    - https://github.com/jsecurity101/MSRPC-to-ATTACK/blob/main/documents/MS-EFSR.md\n" +
+                "    - https://github.com/zeronetworks/rpcfirewall\n" +
+                "    - https://zeronetworks.com/blog/stopping_lateral_movement_via_the_rpc_firewall/\n" +
+                "tags:\n" +
+                "    - attack.defense_evasion\n" +
+                "status: experimental\n" +
+                "author: Sagie Dulce, Dekel Paz\n" +
+                "date: 2022/01/01\n" +
+                "modified: 2022/01/01\n" +
+                "logsource:\n" +
+                "    product: rpc_firewall\n" +
+                "    category: application\n" +
+                "    definition: 'Requirements: install and apply the RPC Firewall to all processes with \"audit:true action:block uuid:df1941c5-fe89-4e79-bf10-463657acf44d or c681d488-d850-11d0-8c52-00c04fd90f7e'\n" +
+                "detection:\n" +
+                "    selection:\n" +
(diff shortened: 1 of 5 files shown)


eirsep:
Thanks for making this change @jowg-amazon.

Have we tested with a sigma rule where condition has `keywords` specified but the log data index has a mapping with only integer or date type fields. If it doesn't match a rule that's expected but the query should not fail with exceptions of any type mismatch.

Can you add a test to verify this scenario doens't fail and simply doesn't generate findings?

src/test/java/org/opensearch/securityanalytics/TestHelpers.java
```diff
@@ -379,6 +379,68 @@ public static String randomRuleWithKeywords() {
...
+                "modified: 2022/01/01\n" +
+                "logsource:\n" +
+                "    product: rpc_firewall\n" +
+                "    category: application\n" +
+                "    definition: 'Requirements: install and apply the RPC Firewall to all processes with \"audit:true action:block uuid:df1941c5-fe89-4e79-bf10-463657acf44d or c681d488-d850-11d0-8c52-00c04fd90f7e'\n" +
+                "detection:\n" +
+                "    selection:\n" +
+                "        EventID: 21\n" +
+                "    keywords:\n" +
+                "        - \"2020-02-04T14:59:39.343541+00:00\"\n" +
```
eirsep, line 437: why are we explicitly adding double quotes to the values in the conditions.. in this the data `2020-02-04T14:59:39.343541+00:00`

why can't it just be 
```
" keywords:\n" +
"        - 2020-02-04T14:59:39.343541+00:00\n" +
```

jowg-amazon: We don't support `2020-02-04T14:59:39.343541+00:00` itself as a field even for non keywords values. I think date types have to be surrounded by either single or double quotes otherwise it throws a parsing exception.

praveensameneni (approved):

eirsep (approved):

PR #788, opened 2023-12-19
Fix duplicate ecs mappings which returns incorrect log index field in mapping view API (#786)

### LINKED ISSUE #786: [BUG] winlog.event_id alias field for windows log type has duplicate mappings which returns incorrect log index field in mapping view API
**What is the bug?**
[`winlog.event_id` ecs field](https://github.com/opensearch-project/security-analytics/blob/main/src/main/resources/OSMapping/windows_logtype.json#L193) has three different raw fields associated with it which can lead to incorrect field being returned by the mappings/view API. This causes an issue when user tries to create a detector with rule(s) that require mapping between the ecs field to a log index field.

For e.g. if the log index has field `EventID`, then the mapping view API should be able to automatically map `winlog.event_id` to `EventID` and send that mapping as part of the response but it ends up sending mapping between `winlog.event_id` to `event_uid` or some other field not present in the index.

1. Create a windows index with `EventID` field
2. Go to create detector page and select the windows index and the windows log type. 
3. In the field mappings section, under the mapped fields tab you will find `winlog.event_id` is not mapped to `EventID`

`winlog.event_id` should be mapped to the field present in the log index.

 Security analytics 2.11

jowg-amazon:
### Description
An ecs field may have multiple raw fields associated with it. Currently, in the `getMappingsView` api, the ecs field will only map to one raw field. The new implementation will automatically map the ecs field to whatever field is present in the index. If multiple raw fields are present in an index, it will automatically map to one of the raw fields and add the other(s) into the `unmappedIndexFields` list.
 
This PR also removes duplicates from the `unmappedFieldAliases` list.

### Issues Resolved
#786 

### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/mapper/MapperService.java
+++ b/src/main/java/org/opensearch/securityanalytics/mapper/MapperService.java
@@ -8,7 +8,6 @@ package org.opensearch.securityanalytics.mapper;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.admin.indices.get.GetIndexRequest;
 import org.opensearch.action.admin.indices.get.GetIndexResponse;
 import org.opensearch.action.admin.indices.mapping.get.GetMappingsRequest;
@@ -509,7 +508,7 @@ public class MapperService {
 
                         // filter out aliases that were included in applyableAliases already
                         List<String> filteredUnmappedFieldAliases = setOfUnmappedFieldAliases.stream()
-                                .filter(e -> applyableAliases.contains(e) == false)
+                                .filter(e -> false == applyableAliases.contains(e))
                                 .collect(Collectors.toList());
 
                         Map<String, Map<String, String>> aliasMappingFields = new HashMap<>();
@@ -518,15 +517,7 @@ public class MapperService {
                             if (allFieldsFromIndex.contains(mapping.getOcsf())) {
                                 aliasMappingFields.put(mapping.getEcs(), Map.of("type", "alias", "path", mapping.getOcsf()));
                             } else if (mapping.getEcs() != null) {
-                                // check if aliasMappingFields already contains a key
-                                if (aliasMappingFields.containsKey(mapping.getEcs())) {
-                                    // if the pathOfApplyableAliases contains the raw field, then override the existing map
-                                    if (pathsOfApplyableAliases.contains(mapping.getRawField())){
-                                        aliasMappingFields.put(mapping.getEcs(), Map.of("type", "alias", "path", mapping.getRawField()));
-                                    }
-                                } else {
-                                    aliasMappingFields.put(mapping.getEcs(), Map.of("type", "alias", "path", mapping.getRawField()));
-                                }
+                                shouldUpdateEcsMappingAndMaybeUpdates(mapping, aliasMappingFields, pathsOfApplyableAliases);
                             } else if (mapping.getEcs() == null) {
                                 aliasMappingFields.put(mapping.getRawField(), Map.of("type", "alias", "path", mapping.getRawField()));
                             }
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/securityanalytics/mapper/MapperService.java
```diff
@@ -501,13 +504,29 @@ public void onResponse(GetMappingsResponse getMappingsResponse) {
...
                             } else if (mapping.getEcs() != null) {
-                                aliasMappingFields.put(mapping.getEcs(), Map.of("type", "alias", "path", mapping.getRawField()));
+                                // check if aliasMappingFields already contains a key
+                                if (aliasMappingFields.containsKey(mapping.getEcs())) {
+                                    // if the pathOfApplyableAliases contains the raw field, then override the existing map
+                                    if (pathsOfApplyableAliases.contains(mapping.getRawField())){
+                                        aliasMappingFields.put(mapping.getEcs(), Map.of("type", "alias", "path", mapping.getRawField()));
+                                    }
+                                } else {
+                                    aliasMappingFields.put(mapping.getEcs(), Map.of("type", "alias", "path", mapping.getRawField()));
```
amsiglan, line 528: This is same as the `if` case. Not sure how it would work differently.

jowg-amazon: The first `if` clause checks if the alias mapping fields already contains the ecs key, if it's true then it checks to make sure that the raw field path that we want to insert is consistent with the one in `pathsOfApplyableAliases`. If it's not consistent, then we skip and do not override the mapping. If the key doesn't exist in the map in the first place, then we directly insert the mapping.

src/main/java/org/opensearch/securityanalytics/mapper/MapperService.java
```diff
@@ -501,13 +504,29 @@ public void onResponse(GetMappingsResponse getMappingsResponse) {
...
+
                         Map<String, Map<String, String>> aliasMappingFields = new HashMap<>();
                         XContentBuilder aliasMappingsObj = XContentFactory.jsonBuilder().startObject();
                         for (LogType.Mapping mapping : requiredFields) {
                             if (allFieldsFromIndex.contains(mapping.getOcsf())) {
                                 aliasMappingFields.put(mapping.getEcs(), Map.of("type", "alias", "path", mapping.getOcsf()));
                             } else if (mapping.getEcs() != null) {
-                                aliasMappingFields.put(mapping.getEcs(), Map.of("type", "alias", "path", mapping.getRawField()));
+                                // check if aliasMappingFields already contains a key
+                                if (aliasMappingFields.containsKey(mapping.getEcs())) {
```
engechas, line 522: Minor: we're already very nested here, we could create a private method to handle this logic rather than nesting further

engechas (approved):

src/main/java/org/opensearch/securityanalytics/mapper/MapperService.java
```diff
@@ -501,13 +504,29 @@ public void onResponse(GetMappingsResponse getMappingsResponse) {
                             }
                         }
 
+                        // turn unmappedFieldAliases into a set to remove duplicates
+                        Set<String> setOfUnmappedFieldAliases = new HashSet<>(unmappedFieldAliases);
+
+                        // filter out aliases that were included in applyableAliases already
+                        List<String> filteredUnmappedFieldAliases = setOfUnmappedFieldAliases.stream()
+                                .filter(e -> applyableAliases.contains(e) == false)
```
eirsep, line 512: nit alway write this condition as `false == <boolean>` for quicker readability and less confusion

eirsep (dismissed):

src/main/java/org/opensearch/securityanalytics/mapper/MapperService.java
jowg-amazon: refactored out the method for better readability

src/main/java/org/opensearch/securityanalytics/mapper/MapperService.java
jowg-amazon: done

riysaxen-amzn (approved):

engechas (approved):

PR #850, opened 2024-02-12
Add execute streaming detectors API

engechas:
### Description
Adds new API for executing streaming detectors and the logic needed to handle these requests 
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
+++ b/src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
@@ -13,6 +13,7 @@ import java.util.function.Supplier;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.cluster.routing.Preference;
+import org.opensearch.commons.alerting.model.Monitor;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.core.action.ActionResponse;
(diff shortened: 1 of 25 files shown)


eirsep:
Can we add a concept of “post-processors” - something that works on the response of the sink call
Use case in mind: I am ingesting data from s3 as source to opensearch as sink, so I make bulk api call to ingest into opensearch. If ingestion succeeds I want to propogate the same events and the success/failure of the api to another processor (after sink) and perform some streaming rule engine process or aggregations etc.
this way we can avoid forking code logic in SAP, Alerting plugins.
Instead invest in normalizing data seen by OSI processor by adding API for mapping OCSF field names to field names in rules from SAP which OSI can leverage

eirsep:
Can we create detectors with aggregation rules on a cluster with these changes?

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
```diff
@@ -248,7 +251,8 @@ public List<Setting<?>> getSettings() {
                 SecurityAnalyticsSettings.CORRELATION_TIME_WINDOW,
                 SecurityAnalyticsSettings.ENABLE_AUTO_CORRELATIONS,
                 SecurityAnalyticsSettings.DEFAULT_MAPPING_SCHEMA,
-                SecurityAnalyticsSettings.ENABLE_WORKFLOW_USAGE
+                SecurityAnalyticsSettings.ENABLE_WORKFLOW_USAGE,
+                SecurityAnalyticsSettings.ENABLE_STREAMING_DETECTORS
```
eirsep, line 255: if i enable the setting > create detector > turn off setting will that detector run normally?

src/main/java/org/opensearch/securityanalytics/model/Detector.java
```diff
@@ -115,13 +113,16 @@ public class Detector implements Writeable, ToXContentObject {
 
     private String findingsIndexPattern;
 
+    private Boolean streamingDetector;
```
eirsep, line 116: if streaming detector is true, why are we not creating workflow with `enabled=false`

eirsep:
can you run `./gradlew integTest` on ensuring security analytics sanity is maintained and there is no regression in correctness because of the new changes

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
engechas: No. Something has to nudge the Workflow entry in `.opendistro-alerting-config` to get the workflow marked as non-streaming.

Or the Workflow entry can be updated manually.

eirsep:
How will pause detector work?

PR #852, opened 2024-02-12
Add an "exists" check for "not" condition in sigma rules

### LINKED ISSUE #854: [BUG] _exists_ check does not exist for "not" condition in sigma rules
**What is the bug?**
A clear and concise description of the bug.
There is a feature gap for customers when configuring a Sigma rule with a not clause. The not clause generates a finding if an ingested document does not match whatever is specified in the detection section. However, this currently also includes cases where that field itself doesn’t exist in the ingested document.

Example of a Pre-Packaged Cloudtrail YAML Rule:

```
id: 8ad1600d-e9dc-4251-b0ee-a65268f29add
logsource:
  product: cloudtrail
title: AWS Root Credentials
description: Detects AWS root account usage
tags:
  - attack.privilege_escalation
  - attack.t1078.004
falsepositives:
  - >-
    AWS Tasks That Require AWS Account Root User Credentials
    https://docs.aws.amazon.com/general/latest/gr/aws_tasks-that-require-root.html
level: medium
status: experimental
references:
  - 'https://docs.aws.amazon.com/IAM/latest/UserGuide/id_root-user.html'
author: vitaliy0x1
detection:
  selection_usertype:
    userIdentity.type: Root
  selection_eventtype:
    eventType: AwsServiceEvent
  condition: selection_usertype and not selection_eventtype
```

Example of an ingested document generating an incorrect finding
```
{
“selection_usertype”: Root
}
```

This above example generates an incorrect finding because of the not selection_eventtype condition since the field eventType is not present in the ingested document.

1. Create a detector with a sigma rule with a `not` condition
2. Insert a document without that field specified in the `not` condition
3. Incorrect finding will be generated

 - OS: 2.11
 - Plugins

Add any other context about the problem.

jowg-amazon:
### Description
Currently the `not` clause generates a finding if an ingested document does not match whatever is specified in the detection section. However, this also includes cases where that field itself doesn’t exist in the ingested document. This PR adds an additional `_exists_` check to the query so that if the field itself is not present in the ingested document, a finding is not generated incorrectly. 

Example: 
OLD: not selection1 -> `NOT field1:val1`
NEW: not selection1 -> `NOT field1:val1 AND _exists_: field1`

Additionally, distributes the `NOT` across parenthesis i.e. applies De Morgan's Theorem.

Example:
OLD: `NOT(a and b)`
NEW: `(NOT a and NOT b)`
 
### Issues Resolved
#854 
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/securityanalytics/rules/backend/QueryBackend.java
```diff
@@ -85,15 +82,15 @@ public List<Object> convertRule(SigmaRule rule) throws SigmaError {
 
                 Object query;
                 if (conditionItem instanceof ConditionAND) {
-                    query = this.convertCondition(new ConditionType(Either.left(AnyOneOf.leftVal((ConditionAND) conditionItem))));
+                    query = this.convertCondition(new ConditionType(Either.left(AnyOneOf.leftVal((ConditionAND) conditionItem))), false, false);
```
eirsep, line 85: why is the `applyDeMorgans` always false? do we even need it??

src/main/java/org/opensearch/securityanalytics/rules/backend/QueryBackend.java
```diff
@@ -113,30 +110,41 @@ public List<Object> convertRule(SigmaRule rule) throws SigmaError {
         return queries;
     }
 
-    public Object convertCondition(ConditionType conditionType) throws SigmaValueError {
+    public Object convertCondition(ConditionType conditionType, boolean isConditionNot, boolean applyDeMorgans) throws SigmaValueError {
```
eirsep, line 113: are both these newly added bool params even necessary?
can't we just set them as necessary for `convertConditionNot`

src/main/java/org/opensearch/securityanalytics/rules/backend/OSQueryBackend.java
```diff
@@ -171,21 +173,29 @@ public Object convertConditionAnd(ConditionAND condition) {
...
-                        converted = this.convertConditionGroup(new ConditionType(Either.right(Either.left(arg.getLeft().getMiddle()))));
+                        converted = this.convertConditionGroup(new ConditionType(Either.right(Either.left(arg.getLeft().getMiddle()))), isConditionNot, applyDeMorgans);
                     } else if (arg.getLeft().isRight()) {
-                        converted = this.convertConditionGroup(new ConditionType(Either.right(Either.right(arg.getLeft().get()))));
+                        converted = this.convertConditionGroup(new ConditionType(Either.right(Either.right(arg.getLeft().get()))), isConditionNot, applyDeMorgans);
                     }
 
                     if (converted != null) {
+                        // if applyDeMorgans is true, then use OR instead of AND
+                        if (applyDeMorgans) {
```
eirsep, line 185: is this ever true?

src/main/java/org/opensearch/securityanalytics/rules/backend/QueryBackend.java
jowg-amazon: `applyDeMorgans` starts off as false but we need it since we are recursively traversing the tree. It gets set to true (ie. we need to apply deMorgans) only if there is an `AND` or `OR` condition when it is an ancestor of a `NOT`. 

Ex. 
1. NOT A OR B -> do not apply deMorgans
2. NOT(A OR B) -> need to apply deMorgans
- This is because when we append the AND _exists_:field to the query, if we don't do deMorgans first but directly apply it directly to the query, an incorrect query would be constructed.

Ex.
Incorrect: NOT(A OR B) -> NOT(A AND _exists_:A OR B AND _exists_:B) -> NOT A OR _exists_:A AND NOT B OR _exists_:B
Correct: NOT(A OR B) -> NOT A AND NOT B -> NOT A AND _exists_:A AND NOT B AND _exists_:B

src/main/java/org/opensearch/securityanalytics/rules/backend/QueryBackend.java
jowg-amazon: see above

eirsep (dismissed):

riysaxen-amzn (dismissed):

riysaxen-amzn (approved):

eirsep (approved):

eirsep:
check buld and test CI

jowg-amazon:
CI's passed successfully

PR #859, opened 2024-02-16
create sap error history index

riysaxen-amzn:
### Description
* This PR creates a SAP history index as per -> https://quip-amazon.com/uc9wARQrnXrw/SAP-History-Index
* Scope is currently limited to create/update detector flows, can be later on expanded to errors related to correlations
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -1573,6 +1576,44 @@ private Map<String, String> mapMonitorIds(List<IndexMonitorResponse> monitorResp
         }
     }
 
+    private void createSAPHistoryIndex(IndexDetectorRequest request, String exception) throws IOException {
```
eirsep, line 1579: plz move this logic to separate class. you can  chekc rule indices handling for reference

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -1522,9 +1519,15 @@ private void onOperation(IndexResponse response, Detector detector) {
...
 
         private void onFailures(Exception t) {
+            log.info("Exception failures while creating detector: {}", t.toString());
             if (counter.compareAndSet(false, true)) {
                 finishHim(null, t);
             }
+            try {
+                createSAPHistoryIndex(request, t.toString());
+            } catch (IOException e) {
+                throw new RuntimeException(e);
```
eirsep, line 1529: 1. plz dont ever throw runtime exceptions.
2.  no need to throw exception at all here imo since this is not a customer facing feature. 
3. plz add error log for failure and continue execution

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -1573,6 +1576,44 @@ private Map<String, String> mapMonitorIds(List<IndexMonitorResponse> monitorResp
...
+        XContentBuilder builder = XContentFactory.jsonBuilder().startObject();
+        builder.field("detectorId", detectorId);
+        builder.field("exception", exception);
+        builder.field("timestamp", timestamp);
+        builder.field("logType", ruleTopic);
+        builder.field("operation", operation);
+        builder.field("user", user);
+        builder.endObject();
+        log.info("Index name is: {}", indexName);
+        IndexRequest indexRequest = new IndexRequest(indexName)
```
eirsep, line 1597: this method is not even creating an index. plz rename method correcty.

eirsep (changes requested):
1. plz add mapping for system index.
2. maintain separate class for system index management. use the right index settings at creation time by referring other system indices' behaviour

praveensameneni (dismissed):

riysaxen-amzn:
> 1. plz add mapping for system index.
> 2. maintain separate class for system index management. use the right index settings at creation time by referring other system indices' behaviour

So, the idea was to create this index whenever we encounter any failure and then upserting it with failure docs. Since this will be an Ops feature, i don't see the utility of creating this index at the beginning. But I see the point since this is a system index it not intended for direct access or modification. So, having a proper system index management for this make sense.

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
riysaxen-amzn: IndexRequest Creates or updates a document in an index. It also creates an index if it doesn't already exists.

PR #860, opened 2024-02-16
Add throw for empty strings in rules with modifier contains, startwith, and endswith

jowg-amazon:
### Description
Currently an array out of bounds exception is thrown when a rule tries to be created on an empty string when modifiers `contains, startswith and endswith` are used. This PR checks for this condition and throws a more descriptive and user friendly error to the front end.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaDetectionItem.java
+++ b/src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaDetectionItem.java
@@ -114,8 +114,8 @@ public class SigmaDetectionItem {
         for (T v: values) {
             SigmaType sigmaType = SigmaTypeFacade.sigmaType(v);
             // throws an error if sigmaType is an empty string and the modifier is "contains" or "startswith" or "endswith"
-            boolean invalidModifier = modifierIds.contains("contains") || modifierIds.contains("startswith") || modifierIds.contains("endswith");
-            if (sigmaType.getClass().equals(SigmaString.class) && v.toString().isEmpty() && invalidModifier) {
+            boolean invalidModifierWithEmptyString = modifierIds.contains("contains") || modifierIds.contains("startswith") || modifierIds.contains("endswith");
+            if (sigmaType.getClass().equals(SigmaString.class) && v.toString().isEmpty() && invalidModifierWithEmptyString) {
                 throw new SigmaValueError("Cannot create rule with empty string and given modifier(s)");
             } else {
                 sigmaTypes.add(sigmaType);
--- a/src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaDetectionItem.java
+++ b/src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaDetectionItem.java
@@ -116,7 +116,7 @@ public class SigmaDetectionItem {
             // throws an error if sigmaType is an empty string and the modifier is "contains" or "startswith" or "endswith"
             boolean invalidModifierWithEmptyString = modifierIds.contains("contains") || modifierIds.contains("startswith") || modifierIds.contains("endswith");
             if (sigmaType.getClass().equals(SigmaString.class) && v.toString().isEmpty() && invalidModifierWithEmptyString) {
-                throw new SigmaValueError("Cannot create rule with empty string and given modifier(s)");
+                throw new SigmaValueError("Cannot create rule with empty string and given modifier(s): " + modifierIds);
             } else {
                 sigmaTypes.add(sigmaType);
             }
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaDetectionItem.java
```diff
@@ -111,7 +112,15 @@ public static <T> SigmaDetectionItem fromMapping(String key, Either<T, List<T>>
 
         List<SigmaType> sigmaTypes = new ArrayList<>();
         for (T v: values) {
-            sigmaTypes.add(SigmaTypeFacade.sigmaType(v));
+            SigmaType sigmaType = SigmaTypeFacade.sigmaType(v);
+            // skips if sigmaType is an empty string and the modifier is "contains" or "startswith" or "endswith"
+            if (sigmaType.getClass().equals(SigmaString.class)) {
+                if (!(v.toString().isEmpty() && (modifierIds.contains("contains") || modifierIds.contains("startswith") || modifierIds.contains("endswith")))) {
+                    sigmaTypes.add(sigmaType);
```
eirsep, line 119: why are we skipping add a type?

jowg-amazon: the `contains`, `startwith`, and `endwith` modifiers try to append a wildcard to the value but if the string is empty, it'll throw an array out of bounds exception which is shown back to the user. This `if` clause will check whether or not the string is empty and if it's one of those modifiers. If it doesn't fulfill that condition then it will add the `sigmaTypes.add(sigmaType)` otherwise it will skip.

jowg-amazon: However, there might be a discrepancy because the rule is allowed to be created but shows an empty string `' '` in the value. The difference this PR makes is that the empty string query isn't generated. As an alternative we can also prevent the rule from being generated all together either from the frontend, backend, or both.

praveensameneni (dismissed):

src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaDetectionItem.java
```diff
@@ -113,11 +113,10 @@ public static <T> SigmaDetectionItem fromMapping(String key, Either<T, List<T>>
...
             SigmaType sigmaType = SigmaTypeFacade.sigmaType(v);
-            // skips if sigmaType is an empty string and the modifier is "contains" or "startswith" or "endswith"
-            if (sigmaType.getClass().equals(SigmaString.class)) {
-                if (!(v.toString().isEmpty() && (modifierIds.contains("contains") || modifierIds.contains("startswith") || modifierIds.contains("endswith")))) {
-                    sigmaTypes.add(sigmaType);
-                }
+            // throws an error if sigmaType is an empty string and the modifier is "contains" or "startswith" or "endswith"
+            boolean invalidModifier = modifierIds.contains("contains") || modifierIds.contains("startswith") || modifierIds.contains("endswith");
+            if (sigmaType.getClass().equals(SigmaString.class) && v.toString().isEmpty() && invalidModifier) {
+                throw new SigmaValueError("Cannot create rule with empty string and given modifier(s)");
```
eirsep, line 119: we need to log the modifiers

jowg-amazon: Added modifiers to the log

eirsep (approved):

src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaDetectionItem.java
```diff
@@ -111,7 +112,14 @@ public static <T> SigmaDetectionItem fromMapping(String key, Either<T, List<T>>
 
         List<SigmaType> sigmaTypes = new ArrayList<>();
         for (T v: values) {
-            sigmaTypes.add(SigmaTypeFacade.sigmaType(v));
+            SigmaType sigmaType = SigmaTypeFacade.sigmaType(v);
+            // throws an error if sigmaType is an empty string and the modifier is "contains" or "startswith" or "endswith"
+            boolean invalidModifierWithEmptyString = modifierIds.contains("contains") || modifierIds.contains("startswith") || modifierIds.contains("endswith");
+            if (sigmaType.getClass().equals(SigmaString.class) && v.toString().isEmpty() && invalidModifierWithEmptyString) {
```
goyamegh, line 118: What will happen to the other values? Should we skip the empty strings and use values which are present in the list?

jowg-amazon: We decided to throw an exception instead of skipping the empty string here because if we allowed the rule creation to go through, the empty string would still appear in the rule itself but could not be turned into a meaningful/correct query. 

Instead we added additional parsing logic on the front end to ensure that empty strings get parsed out when creating a rule from the UI. This PR is an additional check if the rule gets created from the API directly.

eirsep (approved):

goyamegh (approved):

PR #867, opened 2024-02-21
Fix get mappings view API incorrectly returning ecs path

### LINKED ISSUE #866: [BUG] Get Mappings View API incorrectly returns ECS path for OCSF fields
**What is the bug?**
SAP queries are not transformed correctly (with index name and monitor id) when the field in the index is not mapped correctly. There is a bug when a custom rule is created with a raw field name and an index with either raw fields or ocsf fields. The mappings view API returns the ecs name in the alias path instead of the correct ocsf or raw field name even when the ecs format is not present in the index.

1. Create an index with ocsf or raw field field types
2. Create a custom rule with a raw field name(s) in the rule
3. Create a detector with new index and custom rule
4. During detector creation under field mappings see the incorrect ecs path in the data source field for the field specified in the custom rule
5. Can also see this incorrect path in the get Mappings View API for this index/log type

The mappings view API should return the ocsf or the raw field path if a new rule is created using a raw field.

 - OS: 2.12
 - Plugins

Add any other context about the problem.

jowg-amazon:
### Description
Currently the get mappings view api incorrectly returns the ecs path for logs with ocsf or raw field types if a custom rule is created using a raw field. This is because whenever a new rule is created, the rule field(s) gets indexed into `.opensearch-sap-log-types-config` index along with the log_type (ex. below). However, because there are already preset mappings defined for certain log types like `cloudtrail`, the raw_field, ecs, and ocsf mappings are already defined in this index. Then when the get Mappings view API is called, the new document from the custom rule overwrites the existing predefined mapping which causes the ecs format to be incorrectly returned.

```
{
        "_index": ".opensearch-sap-log-types-config",
        "_id": "aws.cloudtrail.event_name|",
        "_score": 1,
        "_source": {
          "raw_field": "aws.cloudtrail.event_name",
          "log_types": [
            "cloudtrail"
          ]
        }
}
```

This PR performs additional logic to check whether or not the raw field from the custom rule already exists in the `existingFieldMappings` (i.e. if this is a known mapping that is predefined in the index). If it is defined, then it doesn't add a the new mapping into the `.opensearch-sap-log-types-config` index. 

The integ test ensures that the mappings view API returns the same output before and after a custom rule with a raw field is created to verify that the correct path is returned.

### Issues Resolved
#866 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/logtype/LogTypeService.java
+++ b/src/main/java/org/opensearch/securityanalytics/logtype/LogTypeService.java
@@ -356,6 +356,9 @@ public class LogTypeService {
                     )) {
                         foundFieldMappingDoc = Optional.of(e);
                     }
+                    // Additional check to see if raw field path + log type combination is already in existingFieldMappings so a new one is not inserted
+                } else if (e.getId().contains(newFieldMapping.getRawField()) && e.getLogTypes().containsAll(newFieldMapping.getLogTypes())) {
+                    foundFieldMappingDoc = Optional.of(e);
                 }
             }
             if (foundFieldMappingDoc.isEmpty()) {
--- a/src/main/java/org/opensearch/securityanalytics/logtype/LogTypeService.java
+++ b/src/main/java/org/opensearch/securityanalytics/logtype/LogTypeService.java
@@ -346,19 +346,27 @@ public class LogTypeService {
         List<FieldMappingDoc> newFieldMappings = new ArrayList<>();
         fieldMappingDocs.forEach( newFieldMapping -> {
             Optional<FieldMappingDoc> foundFieldMappingDoc = Optional.empty();
-            for (FieldMappingDoc e: existingFieldMappings) {
-                if (e.getRawField().equals(newFieldMapping.getRawField())) {
+            for (FieldMappingDoc existingFieldMapping: existingFieldMappings) {
+                if (existingFieldMapping.getRawField().equals(newFieldMapping.getRawField())) {
                     if ((
-                            e.get(defaultSchemaField) != null && newFieldMapping.get(defaultSchemaField) != null &&
-                                    e.get(defaultSchemaField).equals(newFieldMapping.get(defaultSchemaField))
+                            existingFieldMapping.get(defaultSchemaField) != null && newFieldMapping.get(defaultSchemaField) != null &&
+                                    existingFieldMapping.get(defaultSchemaField).equals(newFieldMapping.get(defaultSchemaField))
                     ) || (
-                            e.get(defaultSchemaField) == null && newFieldMapping.get(defaultSchemaField) == null
+                            existingFieldMapping.get(defaultSchemaField) == null && newFieldMapping.get(defaultSchemaField) == null
                     )) {
-                        foundFieldMappingDoc = Optional.of(e);
+                        foundFieldMappingDoc = Optional.of(existingFieldMapping);
                     }
-                    // Additional check to see if raw field path + log type combination is already in existingFieldMappings so a new one is not inserted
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/securityanalytics/logtype/LogTypeService.java
```diff
@@ -356,6 +356,9 @@ private List<FieldMappingDoc> mergeFieldMappings(List<FieldMappingDoc> existingF
                     )) {
                         foundFieldMappingDoc = Optional.of(e);
                     }
+                    // Additional check to see if raw field path + log type combination is already in existingFieldMappings so a new one is not inserted
+                } else if (e.getId().contains(newFieldMapping.getRawField()) && e.getLogTypes().containsAll(newFieldMapping.getLogTypes())) {
```
engechas, line 360: Wasn't added by you so optional - Maybe `existingFieldMapping` instead of `e`. `e` reads as Exception to me

engechas: Is there a possibility of false positives here? Say I have a field with name `field2` and then another field with name `field`. If `field2` is evaluated first, the `contains` call would match for `field` even though it's an entry for `field2`

jowg-amazon: Yeah I agree `e` is a bit confusing. I changed it to `existingFieldMapping`. Also yes that false positive is a possibility. I changed the logic so that we grab the right side of the ID of the doc which corresponds to the ecs field name when we load in our predefined mappings. That way we can do an equals check instead to ensure that there are no duplicates. 

ex. of document from pre-defined mapping
```
{
        "_index": ".opensearch-sap-log-types-config",
        "_id": "eventName|aws.cloudtrail.event_name",
        "_score": 1,
        "_source": {
          "raw_field": "eventName",
          "ecs": "aws.cloudtrail.event_name",
          "ocsf": "api.operation",
          "log_types": [
            "s3",
            "cloudtrail"
          ]
        }
      }
```

src/main/java/org/opensearch/securityanalytics/logtype/LogTypeService.java
```diff
@@ -346,15 +346,26 @@ private List<FieldMappingDoc> mergeFieldMappings(List<FieldMappingDoc> existingF
...
                     )) {
-                        foundFieldMappingDoc = Optional.of(e);
+                        foundFieldMappingDoc = Optional.of(existingFieldMapping);
+                    }
+                    // Grabs the right side of the ID with "|" as the delimiter if present representing the ecs field from predefined mappings
+                    // Additional check to see if raw field path + log type combination is already in existingFieldMappings so a new one is not indexed
+                } else {
+                    String id = existingFieldMapping.getId();
+                    int indexOfPipe = id.indexOf("|");
+                    if (indexOfPipe != -1) {
```
engechas, line 364: Is this backwards compatible? Wondering if we will have duplicate mappings for a mapping that was generated before the pipe system was introduced (say an upgrade from an older to the current version).

I am also not very familiar with this part of the code though, so if that is not a concern, then I am good to approve.

eirsep:
@sbcd90 can we change logic to remove the code for adding any new field into this index at all during custom rule creation?

sbcd90 (approved):

src/main/java/org/opensearch/securityanalytics/logtype/LogTypeService.java
jowg-amazon: Backwards compatibility should be fine because the mapping should not have been inserted in the first place but now we're ensuring that it doesn't get generated. If they encounter this bug though I think they may need to delete the log-type-config index so that we can automatically regenerate it with the correct mappings.

PR #874, opened 2024-03-01
Fixing hanging tasks for correlations

goyamegh:
### Description
While dealing with correlations, we noticed unexpectedly high resource consumption in CPU and JVM, resulting in node drops. Upon closer examination, we found a subset of subscribe/findings tasks consistently in a hanging state throughout the data node's lifecycle until it either dropped or required manual restart. This information can be corroborated through the _cat/tasks API. 

The root cause of these hanging tasks was the way action listeners handle exceptions today, causing the parent listener to not close in some cases, if the exceptions are not caught correctly.

In this specific case, the exception causing the contention of resources was:
`
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
        at org.opensearch.securityanalytics.correlation.VectorEmbeddingsEngine$1$1.onResponse(VectorEmbeddingsEngine.java:142)
        at org.opensearch.securityanalytics.correlation.VectorEmbeddingsEngine$1$1.onResponse(VectorEmbeddingsEngine.java:115)
        at org.opensearch.action.support.TransportAction$1.onResponse(TransportAction.java:113)
        at org.opensearch.action.support.TransportAction$1.onResponse(TransportAction.java:107)
        at org.opensearch.action.search.TransportMultiSearchAction$1.finish(TransportMultiSearchAction.java:209)
        at org.opensearch.action.search.TransportMultiSearchAction$1.handleResponse(TransportMultiSearchAction.java:195)
        at org.opensearch.action.search.TransportMultiSearchAction$1.onResponse(TransportMultiSearchAction.java:183)
        at org.opensearch.action.search.TransportMultiSearchAction$1.onResponse(TransportMultiSearchAction.java:180)
        at org.opensearch.action.support.TransportAction$1.onResponse(TransportAction.java:113)
        at org.opensearch.action.support.TransportAction$1.onResponse(TransportAction.java:107)
        at org.opensearch.performanceanalyzer.action.PerformanceAnalyzerActionListener.onResponse(PerformanceAnalyzerActionListener.java:55)
        at org.opensearch.action.support.TimeoutTaskCancellationUtility$TimeoutRunnableListener.onResponse(TimeoutTaskCancellationUtility.java:132)
`

This will PR will be followed up by two more solutions:
1. Fix the working of child action listeners across Security Analytics to handle the closing of parent listener tasks correctly
2. Analyze and see how the methods 'getHits().length`and `getTotalHits().value` are different and if we need to modify the way the search requests have been created in order to restore this usage in future.

### Issues Resolved
This PR aims to fix the current exception being thrown by using `getHits().length` instead of `getTotalHits().value` as a mitigation to avoid any tasks to hang.
 
### Check List
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
+++ b/src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
@@ -7,6 +7,7 @@ package org.opensearch.securityanalytics.correlation;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.OpenSearchStatusException;
+import org.opensearch.ResourceNotFoundException;
 import org.opensearch.cluster.routing.Preference;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.action.bulk.BulkRequest;
@@ -84,6 +85,11 @@ public class VectorEmbeddingsEngine {
                     correlateFindingAction.onFailures(new OpenSearchStatusException("Search request timed out", RestStatus.REQUEST_TIMEOUT));
                 }
 
+                if (response.getHits().getHits().length == 0) {
+                    correlateFindingAction.onFailures(
+                            new ResourceNotFoundException("Failed to find hits in metadata index for finding id {}", finding.getId()));
+                }
+
                 Map<String, Object> hitSource = response.getHits().getHits()[0].getSourceAsMap();
                 long counter = Long.parseLong(hitSource.get("counter").toString());
 
@@ -125,7 +131,7 @@ public class VectorEmbeddingsEngine {
                                 continue;
                             }
 
-                            long totalHits = response.getResponse().getHits().getTotalHits().value;
+                            long totalHits = response.getResponse().getHits().getHits().length;
                             totalNeighbors += totalHits;
 
                             for (int idx = 0; idx < totalHits; ++idx) {
--- a/src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
+++ b/src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
@@ -8,6 +8,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.lucene.search.join.ScoreMode;
 import org.opensearch.OpenSearchStatusException;
+import org.opensearch.ResourceNotFoundException;
 import org.opensearch.cluster.routing.Preference;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.action.ActionRequest;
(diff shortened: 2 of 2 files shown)


src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
```diff
@@ -125,7 +131,7 @@ public void onResponse(MultiSearchResponse items) {
                                 continue;
                             }
 
-                            long totalHits = response.getResponse().getHits().getTotalHits().value;
+                            long totalHits = response.getResponse().getHits().getHits().length;
```
engechas, line 134: What's the rationale to swap from the total hits value to the hits array length?

engechas (approved):

src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
goyamegh: We are seeing inconsistent results from search response where the getTotalHits().value is non zero, but the actual response array has no objects. The idea is to use the same array for both, validation and then fetching data.

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -517,6 +518,11 @@ public void onFailure(Exception e) {
                 client.search(searchRequest, new ActionListener<>() {
                     @Override
                     public void onResponse(SearchResponse response) {
+                        if (response.getHits().getHits().length == 0) {
```
eirsep, line 521: is not getting any hits really an exception? i thought if `totalHits>0 
 && response.getHits().getHits().length == 0` only then its an exceptional case

goyamegh: The search request expects the result to have one doc (size is set to 1), we can anyway get only two responses from the hits array length (0 or 1). Total hits isn't a required check for this invocation.

eirsep (approved):

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -655,6 +661,7 @@ public void onOperation() {
         }
 
         public void onFailures(Exception t) {
+            log.error("Exception occurred while processing correlations", t);
```
eirsep, line 664: redundant log line. the correct message should be logged at caller.

goyamegh: This line will be the only place we log errors before throwing the exception out to Alerting, I made sure of this while testing for most happy cases, will follow up and remove them in future PRs.

engechas (approved):

PR #880, opened 2024-03-04
Refactor invocation of Action listeners in correlations

### LINKED ISSUE #879: [FEATURE] Enhance System Performance to address Hanging Tasks by Optimizing Correlations
Inefficient correlation processes within the system are leading to hanging tasks, causing performance degradation and impacting system reliability. These hanging tasks not only disrupt workflow but also hinder the timely processing of critical operations, resulting in decreased productivity and even node drops as a result. There is a pressing need to optimize the correlation logic and implement robust measures to mitigate hanging tasks, thereby enhancing system performance and reliability.

The existing correlation logic in the system fails to effectively manage parent tasks associated with action listeners used for correlations. Consequently, unresolved parent tasks persist, impeding workflow continuity and hindering essential system operations. To address this issue and enhance task management efficiency, a comprehensive refactor of the correlation logic is imperative to enable the timely closure of parent tasks linked to action listeners.

One such example of exceptions not floating back was addressed in: https://github.com/opensearch-project/security-analytics/pull/874

goyamegh:
### Description
This PR is intended to fix the hanging tasks observed in _cat/tasks by refactoring the correlation workflow to ensure timely closure of parent action listeners upon successful completion or encountering exceptions, and consolidating exception handling logic into a centralized function. The aim is to optimize task management efficiency and enhance the overall reliability of our system.

This logic has been tested against a high indexing workload ( approx. 1 M docs/ minute) where the issue was observed prominently in a cluster of 3 or more data nodes, and generating findings with the help of a Cloudtrail logs detector running all 32 pre-packaged rules at a frequency of 1 minute. Further, the correlations were generated with the help of a single rule on the same log type for testing, where the findings are generated at a rate of 1~2k per minute.
 
### Issues Resolved
#879
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
+++ b/src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
@@ -91,7 +91,7 @@ public class JoinEngine {
                 onAutoCorrelations(detector, finding, Map.of());
             }
         } catch (IOException ex) {
-            correlateFindingAction.onFailures(ex);
+            onFailure(ex);
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
```diff
@@ -231,39 +215,34 @@ private void onAutoCorrelations(Detector detector, Finding finding, Map<String,
...
+                onFailure(new OpenSearchStatusException("Search request timed out", RestStatus.REQUEST_TIMEOUT));
             }
 
-            @Override
-            public void onFailure(Exception e) {
-                getValidDocuments(detectorType, indices, List.of(), List.of(), autoCorrelations);
+            Iterator<SearchHit> hits = response.getHits().iterator();
+            List<CorrelationRule> correlationRules = new ArrayList<>();
+            while (hits.hasNext()) {
+                try {
```
eirsep, line 226: nit: Try-catch is redundant as we are using ActionListener.wrap() which catches generic Exception

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -253,39 +230,31 @@ void start() {
...
+                    if (response.isTimedOut()) {
+                        onFailures(new OpenSearchStatusException("Search request timed out", RestStatus.REQUEST_TIMEOUT));
                     }
 
-                    @Override
-                    public void onFailure(Exception e) {
-                        onFailures(e);
+                    SearchHits hits = response.getHits();
+                    // Detectors Index hits count could be more even if we fetch one
+                    if (hits.getTotalHits().value >= 1 && hits.getHits().length > 0) {
```
eirsep, line 240: this check looks different in every place. 
Why cant we just iterate hits with a for loop and not do any of these checks

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -253,39 +230,31 @@ void start() {
...
+                                    LoggingDeprecationHandler.INSTANCE, hit.getSourceAsString()
+                            );
+                            Detector detector = Detector.docParse(xcp, hit.getId(), hit.getVersion());
+                            joinEngine.onSearchDetectorResponse(detector, finding);
+                        } catch (IOException e) {
+                            log.error("IOException for request {}", searchRequest.toString(), e);
+                            onFailures(e);
+                        }
+                    } else {
+                        onFailures(new OpenSearchStatusException("detector not found given monitor id", RestStatus.INTERNAL_SERVER_ERROR));
```
eirsep, line 255: log the monitor id

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -326,327 +287,205 @@ public void onFailure(Exception e) {
...
-                                                                    vectorEmbeddingsEngine.insertCorrelatedFindings(detectorType, request.getFinding(), correlatedFinding.getKey(), correlatedFinding.getValue(),
-                                                                            timestampFeature, correlationRules, logTypes);
-                                                                }
-                                                            } else {
-                                                                vectorEmbeddingsEngine.insertOrphanFindings(detectorType, orphanFinding, timestampFeature, logTypes);
-                                                            }
-                                                        }
+                            correlationIndices.setupCorrelationIndex(indexTimeout, setupTimestamp, ActionListener.wrap(bulkResponse -> {
+                                if (bulkResponse.hasFailures()) {
+                                    log.error(new OpenSearchStatusException(bulkResponse.toString(), RestStatus.INTERNAL_SERVER_ERROR));
```
eirsep, line 296: onfailure()?

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -326,327 +287,205 @@ public void onFailure(Exception e) {
...
+                                SearchRequest searchMetadataIndexRequest = getSearchMetadataIndexRequest();
 
-                                            @Override
-                                            public void onFailure(Exception e) {
-                                                onFailures(e);
-                                            }
-                                        });
-                                    }
+                                client.search(searchMetadataIndexRequest, ActionListener.wrap(searchMetadataResponse -> {
+                                    String id = searchMetadataResponse.getHits().getHits()[0].getId();
```
eirsep, line 303: size check on hits array?

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -326,327 +287,205 @@ public void onFailure(Exception e) {
...
+                                                            vectorEmbeddingsEngine.insertCorrelatedFindings(detectorType, request.getFinding(), correlatedFinding.getKey(), correlatedFinding.getValue(),
+                                                                    Long.valueOf(CorrelationIndices.FIXED_HISTORICAL_INTERVAL / 1000L).floatValue(), correlationRules, logTypes);
+                                                        }
+                                                    } else {
+                                                        vectorEmbeddingsEngine.insertOrphanFindings(detectorType, orphanFinding, Long.valueOf(CorrelationIndices.FIXED_HISTORICAL_INTERVAL / 1000L).floatValue(), logTypes);
+                                                    }
+                                                }, this::onFailures));
+                                            }, this::onFailures));
+                                        } catch (Exception ex) {
+                                            onFailures(ex);
```
eirsep, line 342: log exception

eirsep (changes requested):

src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
```diff
@@ -231,39 +215,34 @@ private void onAutoCorrelations(Detector detector, Finding finding, Map<String,
...
-                        correlationRules.add(rule);
-                    } catch (IOException e) {
-                        correlateFindingAction.onFailures(e);
-                    }
-                }
-
-                getValidDocuments(detectorType, indices, correlationRules, relatedDocIds, autoCorrelations);
+        client.search(searchRequest, ActionListener.wrap(response -> {
+            if (response.isTimedOut()) {
+                onFailure(new OpenSearchStatusException("Search request timed out", RestStatus.REQUEST_TIMEOUT));
```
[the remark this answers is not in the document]
goyamegh: onFailure() will inadvertently throw the exception in finishHim() function in TransportCorrelateFindingsAction.java

src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
```diff
@@ -231,39 +215,34 @@ private void onAutoCorrelations(Detector detector, Finding finding, Map<String,
...
+                    SearchHit hit = hits.next();
+
+                    XContentParser xcp = XContentType.JSON.xContent().createParser(
+                            xContentRegistry,
+                            LoggingDeprecationHandler.INSTANCE,
+                            hit.getSourceAsString());
+
+                    CorrelationRule rule = CorrelationRule.parse(xcp, hit.getId(), hit.getVersion());
+                    correlationRules.add(rule);
+                } catch (IOException e) {
```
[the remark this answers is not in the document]
goyamegh: onFailure() calls TransportCorrelateFindings.onFailures(), which will make sure that we log this, and that too once in the lifetime of the task.

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
goyamegh: Same comment as before

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
goyamegh: We haven't been following this practice for most bulk requests. But, I checked that the setupCorrelationIndex() is simply indexing two docs, which are later needed as well. Modifying this behavior to fail here itself. @sbcd90 please  verify. This is one of the cases where we observed exceptions on, metadata index not having the right docs.

Rectifying this for other calls too.

src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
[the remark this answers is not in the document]
eirsep: we don't need to log at generic failure handling code blocks. we should log in the method so that we are able to communicate where the exception came from or pass a custom error log message to on failure (i would prefer second approach but that would require a lot more change in impl. so was suggesting simple error log. )

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
goyamegh: Added.

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
goyamegh: I think the problem is that the usage is different everywhere. This usage only expects a single doc and probably why it was written like this. I can take a quick pass at all usages of getTotalHits() in a follow-up and fix this everywhere to make it consistent. 

Removing the check on total hits entirely.

src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
```diff
@@ -306,84 +285,72 @@ private void getValidDocuments(String detectorType, List<String> indices, List<C
...
+                        log.info(response.getFailureMessage());
+                        continue;
+                    }
 
-                        if (response.getResponse().getHits().getTotalHits().value > 0L) {
-                            filteredCorrelationRules.add(new FilteredCorrelationRule(validCorrelationRules.get(idx),
-                                    response.getResponse().getHits().getHits(), validFields.get(idx)));
-                        }
-                        ++idx;
+                    if (response.getResponse().getHits().getTotalHits().value > 0L) {
```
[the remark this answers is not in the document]
goyamegh: Corrected.

src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
```diff
@@ -61,208 +58,203 @@ public VectorEmbeddingsEngine(Client client, TimeValue indexTimeout, long corrTi
...
-                        long prevCounter = -1L;
-                        long totalNeighbors = 0L;
-                        for (MultiSearchResponse.Item response: responses) {
-                            if (response.isFailure()) {
-                                log.info(response.getFailureMessage());
-                                continue;
-                            }
+        client.search(searchRequest, ActionListener.wrap(response -> {
+            if (response.isTimedOut()) {
+                onFailure(new OpenSearchStatusException("Search request timed out", RestStatus.REQUEST_TIMEOUT));
```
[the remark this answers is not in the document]
goyamegh: It'll happen in finishHim().

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -134,67 +135,43 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Subscr
...
-                        public void onResponse(CreateIndexResponse response) {
-                            if (response.isAcknowledged()) {
-                                IndexUtils.correlationIndexUpdated();
-                                if (IndexUtils.correlationIndexUpdated) {
-                                    IndexUtils.lastUpdatedCorrelationHistoryIndex = IndexUtils.getIndexNameWithAlias(
-                                            clusterService.state(),
-                                            CorrelationIndices.CORRELATION_HISTORY_WRITE_INDEX
-                                    );
-                                }
+                    AsyncCorrelateFindingAction correlateFindingAction = new AsyncCorrelateFindingAction(task, transformedRequest, actionListener);
```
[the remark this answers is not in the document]
goyamegh: Not unintentional. We will use a single onFailures() function across the entire flow. The idea to move it before if condition is to use the same function.

src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
```diff
@@ -231,39 +215,34 @@ private void onAutoCorrelations(Detector detector, Finding finding, Map<String,
...
+                    CorrelationRule rule = CorrelationRule.parse(xcp, hit.getId(), hit.getVersion());
+                    correlationRules.add(rule);
+                } catch (IOException e) {
+                    onFailure(e);
+                }
             }
-        });
+            getValidDocuments(detectorType, indices, correlationRules, relatedDocIds, autoCorrelations);
+        }, e -> {
+            log.error("[CORRELATIONS] Exception encountered while searching correlation rule index for finding id {}",
```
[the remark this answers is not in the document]
goyamegh: Makes sense. Added in getTimesStampFeature() too.

src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
goyamegh: Removed.

src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
```diff
@@ -231,39 +215,34 @@ private void onAutoCorrelations(Detector detector, Finding finding, Map<String,
...
+
+                    XContentParser xcp = XContentType.JSON.xContent().createParser(
+                            xContentRegistry,
+                            LoggingDeprecationHandler.INSTANCE,
+                            hit.getSourceAsString());
+
+                    CorrelationRule rule = CorrelationRule.parse(xcp, hit.getId(), hit.getVersion());
+                    correlationRules.add(rule);
+                } catch (IOException e) {
+                    onFailure(e);
```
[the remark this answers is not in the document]
goyamegh: Umm, that's not true. Once an exception is raised, the loop will break for this thread. Since we are closing the parent listener, so we should be good.

Anyway, I removed it.

src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
```diff
@@ -61,208 +58,203 @@ public VectorEmbeddingsEngine(Client client, TimeValue indexTimeout, long corrTi
...
+            }
 
-                            long totalHits = response.getResponse().getHits().getTotalHits().value;
-                            totalNeighbors += totalHits;
+            if (response.getHits().getHits().length == 0) {
+                onFailure(
```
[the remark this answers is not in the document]
goyamegh: Seems like some confusion. The same logic is used twice. No change in logic was done here. 
Check these lines for reference:
1. https://github.com/opensearch-project/security-analytics/blob/main/src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java#L88
2. https://github.com/opensearch-project/security-analytics/blob/main/src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java#L134

src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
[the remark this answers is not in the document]
goyamegh: Makes sense. For now, I have logged the following information for now:
1. Exception and stack trace
2. Monitor and finding id of that request
3. Trace should tell us where the exception originated from.

I'm not reverting this for now, as we are missing logging completely at this point. As a long term, I created a git hub issue to take this up: https://github.com/opensearch-project/security-analytics/issues/883

src/main/java/org/opensearch/securityanalytics/correlation/JoinEngine.java
```diff
@@ -231,39 +215,34 @@ private void onAutoCorrelations(Detector detector, Finding finding, Map<String,
...
 
-            @Override
-            public void onFailure(Exception e) {
+                CorrelationRule rule = CorrelationRule.parse(xcp, hit.getId(), hit.getVersion());
+                correlationRules.add(rule);
+            }
+            getValidDocuments(detectorType, indices, correlationRules, relatedDocIds, autoCorrelations);
+        }, e -> {
+            try {
+                log.error("[CORRELATIONS] Exception encountered while searching correlation rule index for finding id {}",
```
engechas, line 239: Minor: we should remove the `[CORRELATIONS] prefix`

engechas (approved):

eirsep (approved):

PR #893, opened 2024-03-07
Add search request timeouts for correlations workflows

### LINKED ISSUE #879: [FEATURE] Enhance System Performance to address Hanging Tasks by Optimizing Correlations

goyamegh:
### Description
Add search request timeouts to close hanging tasks for correlations waiting on searches.
 
### Issues Resolved
#879
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
+++ b/src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
@@ -195,6 +195,14 @@ public class VectorEmbeddingsEngine {
     }
 
     public void insertOrphanFindings(String detectorType, Finding finding, float timestampFeature, Map<String, CustomLogType> logTypes) {
+        if (logTypes.get(detectorType) == null ) {
+            log.error("[PERF-DEBUG] insertOrphanFindings detector type {} {}", detectorType, finding.getId());
+            for (String key : logTypes.keySet()) {
+                log.error("[PERF-DEBUG] keys {}", key);
+            }
+            onFailure(new OpenSearchStatusException("insertOrphanFindings null log types for detector type: " + detectorType, RestStatus.INTERNAL_SERVER_ERROR));
+        }
+
         SearchRequest searchRequest = getSearchMetadataIndexRequest(detectorType, finding, logTypes);
         Map<String, Object> tags = logTypes.get(detectorType).getTags();
         String correlationId = tags.get("correlation_id").toString();
@@ -407,6 +415,8 @@ public class VectorEmbeddingsEngine {
                                             } catch (Exception ex) {
                                                 onFailure(ex);
                                             }
+                                        } else {
+                                            onFailure(new OpenSearchStatusException("Indexing failed", RestStatus.INTERNAL_SERVER_ERROR));
                                         }
                                     }, this::onFailure));
                                 } catch (Exception ex) {
--- a/src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
+++ b/src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
@@ -32,6 +32,7 @@ import org.opensearch.securityanalytics.model.CustomLogType;
 import org.opensearch.securityanalytics.transport.TransportCorrelateFindingAction;
 import org.opensearch.securityanalytics.util.CorrelationIndices;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


engechas (dismissed):

src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
```diff
@@ -195,6 +196,14 @@ public void insertCorrelatedFindings(String detectorType, Finding finding, Strin
     }
 
     public void insertOrphanFindings(String detectorType, Finding finding, float timestampFeature, Map<String, CustomLogType> logTypes) {
+        if (logTypes.get(detectorType) == null ) {
+            log.debug("[PERF-DEBUG] insertOrphanFindings detector type {} {}", detectorType, finding.getId());
+            for (String key : logTypes.keySet()) {
+                log.debug("[PERF-DEBUG] keys {}", key);
```
eirsep, line 202: Plz remove perf_debug prefix

src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
```diff
@@ -407,6 +417,8 @@ public void insertOrphanFindings(String detectorType, Finding finding, float tim
                                             } catch (Exception ex) {
                                                 onFailure(ex);
                                             }
+                                        } else {
+                                            onFailure(new OpenSearchStatusException("Indexing failed", RestStatus.INTERNAL_SERVER_ERROR));
```
eirsep, line 421: Can log message convey that search timed out hence indexing failed

src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
```diff
@@ -94,6 +94,7 @@ public void insertCorrelatedFindings(String detectorType, Finding finding, Strin
                 request.indices(CorrelationIndices.CORRELATION_HISTORY_INDEX_PATTERN_REGEXP);
                 request.source(searchSourceBuilder);
                 request.preference(Preference.PRIMARY_FIRST.type());
+                request.setCancelAfterTimeInterval(TimeValue.timeValueSeconds(10L));
```
eirsep, line 97: Why such an aggressive timeout

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -364,54 +366,49 @@ public void getTimestampFeature(String detectorType, Map<String, List<String>> c
                         if (response.getHits().getHits().length == 0) {
                             onFailures(new ResourceNotFoundException(
                                     "Failed to find hits in metadata index for finding id {}", request.getFinding().getId()));
-                        }
-
-                        String id = response.getHits().getHits()[0].getId();
-                        Map<String, Object> hitSource = response.getHits().getHits()[0].getSourceAsMap();
-                        long scoreTimestamp = (long) hitSource.get("scoreTimestamp");
+                        } else {
+                            String id = response.getHits().getHits()[0].getId();
```
eirsep, line 370: Empty check?(i see u didnt change the code but better to add)

eirsep (changes requested):
Lgtm 
Minor comments

src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
goyamegh: Fixed logging.

src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
goyamegh: Fixed the messaging to include the response itself as timeout can be one of the responses.

src/main/java/org/opensearch/securityanalytics/correlation/VectorEmbeddingsEngine.java
goyamegh: We were seeing correlations tasks hanging around due to searches. Since these are system indices, we have configured a standard 10 second timeout for now, but we can discuss and make it configurable in future.

src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
goyamegh: I added this check intentionally as I was observing that the flow was continuing even when we throw the exception in the if block above. So, putting this in the else block will avoid the ArrayOutOfBoundsException then.

eirsep (approved):
typically timeout should be a setting but we can do a quick follow up for that

engechas (approved):

PR #972, opened 2024-04-05
[BUG] Changes doc level query name field from id to rule name and adds validation

jowg-amazon:
### Description
Changes doc level query name field from id to rule name.

Adds validation check for Sigma rule title during rule creation/updation. Sigma rule title is restricted so that it must be 1 - 256 chars as per Sigma rule specifications.

https://github.com/SigmaHQ/sigma/wiki/Specification/10f4159a3ee87c2f322a137bad6cdad313a169ab#title

### Issues Resolved
PR related to:
- https://github.com/opensearch-project/common-utils/pull/630
- https://github.com/opensearch-project/alerting/pull/1506
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaRule.java
```diff
@@ -179,6 +181,20 @@ public static SigmaRule fromYaml(String rule, boolean collectErrors) throws Sigm
         return fromDict(ruleMap, collectErrors);
     }
 
+    public static void validateSigmaRuleTitle(String title, List<SigmaError> errors)
+    {
+        // allowed characters [- : , ( ) [ ] ' _]
+        String allowedChars = "-:,\\(\\)\\[\\]\'_";
+        // regex to restrict string to alphanumeric and allowed chars, must be between 0 - 256 characters
+        String regex = "[\\w\\s" + Pattern.quote(allowedChars) + "]{0,256}";
```
engechas, line 189: Minor: since we use this pattern in both SA and alerting, we should move it to common utils so that it can be defined in one place. It would be easy to miss updating this pattern in either alerting or SA if it changes in the future

engechas (dismissed):

src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaRule.java
jowg-amazon: Moved the validation to common utils inside `ValidationHelpers.kt`

engechas (dismissed):

AWSHurneyt (approved):

engechas (approved):

riysaxen-amzn:
i see many tests are failing.. can we check that

jowg-amazon:
> i see many tests are failing.. can we check that

They're failing because of common util changes, will rerun the CIs

src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaRule.java
```diff
@@ -105,6 +106,17 @@ protected static SigmaRule fromDict(Map<String, Object> rule, boolean collectErr
             ruleId = null;
         }
 
+        String title;
```
riysaxen-amzn, line 109: better way to write

```
String title = rule.containsKey("title") ? rule.get("title").toString() : "";
validateTitle(title);

// Validation method
private void validateTitle(String title) {
    if (!title.matches("^.{1,256}$")) {
        errors.add(new SigmaTitleError("Sigma rule title can be max 256 characters"));
    }
}

```

riysaxen-amzn (approved):
Approved with a minor comment. Also, let's make sure CI passes atleast once even if the tests are flaky

src/main/java/org/opensearch/securityanalytics/rules/objects/SigmaRule.java
jowg-amazon: Can do this in a follow up PR, I did it this way to remain consistent with how we do the other validation checks right now.

engechas (approved):

PR #1048, opened 2024-05-29
Draft of IOC service

AWSHurneyt:
### Description
Implemented draft of IocService.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
```diff
@@ -111,12 +111,16 @@ public class SecurityAnalyticsPlugin extends Plugin implements ActionPlugin, Map
...
     public static final String JOB_INDEX_NAME = ".opensearch-sap--job";
     public static final Map<String, Object> TIF_JOB_INDEX_SETTING = Map.of(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1, IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, "0-all", IndexMetadata.SETTING_INDEX_HIDDEN, true);
-    public static final String IOC_INDEX_NAME_BASE = ".opensearch-sap-ioc";
+    public static final String IOC_INDEX_NAME_BASE = ".opensearch-sap-iocs";
     public static final String IOC_ALL_INDEX_PATTERN = IOC_INDEX_NAME_BASE + "-*";
     public static final String IOC_DOMAIN_INDEX_NAME = IOC_INDEX_NAME_BASE + IocDao.IocType.DOMAIN.name().toLowerCase(Locale.ROOT);
     public static final String IOC_HASH_INDEX_NAME = IOC_INDEX_NAME_BASE + IocDao.IocType.HASH.name().toLowerCase(Locale.ROOT);
     public static final String IOC_IP_INDEX_NAME = IOC_INDEX_NAME_BASE + IocDao.IocType.IP.name().toLowerCase(Locale.ROOT);
 
+    // TODO hurneyt IOC system index names should be refactored to include the timestamp instead of the IOC type
```
eirsep, line 120: plz also incorporate feed id in index name

AWSHurneyt: Good callout, will do.

src/test/java/org/opensearch/securityanalytics/services/IocServiceIT.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static org.opensearch.securityanalytics.SecurityAnalyticsPlugin.IOC_ALL_INDEX_PATTERN;
+import static org.opensearch.securityanalytics.SecurityAnalyticsPlugin.IOC_IP_INDEX_NAME;
+
+public class IocServiceIT extends OpenSearchIntegTestCase {
```
eirsep, line 39: thanks for adding integ tests for service! :D

src/main/java/org/opensearch/securityanalytics/services/IocService.java
```diff
@@ -0,0 +1,131 @@
...
+import org.opensearch.securityanalytics.threatIntel.common.StashedThreadContext;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.opensearch.securityanalytics.SecurityAnalyticsPlugin.IOC_INDEX_NAME_BASE;
+
+/**
+ * IOC Service implements operations that interact with retrieving IOCs from data sources,
```
eirsep, line 39: can we have feed/iocs downloader seperate from the class that is indexing the iocs into system index.

src/main/java/org/opensearch/securityanalytics/model/IocDao.java
```diff
@@ -21,6 +21,7 @@
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
```
eirsep, line 24: NIT: can we rename class to `Ioc`?  to keep consistence with model naming like Monitor.kt or Detector.java

src/main/java/org/opensearch/securityanalytics/services/IocService.java
```diff
@@ -0,0 +1,131 @@
...
+        this.clusterService = clusterService;
+    }
+
+    /**
+     * Checks whether the [IOC_INDEX_NAME_BASE]-related index exists.
+     * @param index The index to evaluate.
+     * @return TRUE if the index is an IOC-related system index, and exists; else returns FALSE.
+     */
+    public boolean hasIocSystemIndex(String index) {
+        return index.startsWith(IOC_INDEX_NAME_BASE) && this.clusterService.state().routingTable().hasIndex(index);
```
eirsep, line 58: why do we have to do `index.startsWith(IOC_INDEX_NAME_BASE)` and why is method public?

src/main/java/org/opensearch/securityanalytics/services/IocService.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public void initSystemIndexes(String index, ActionListener<FetchIocsActionResponse> listener) {
```
eirsep, line 61: what is FetchIocsActionResponse? why is the invoker of init system indices method listening for FetchIocsActionResponse ?

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
```diff
@@ -111,12 +111,16 @@ public class SecurityAnalyticsPlugin extends Plugin implements ActionPlugin, Map
...
+    public static final String IOC_HISTORY_WRITE_INDEX_ALIAS = IOC_INDEX_NAME_BASE + "-history-write";
+    public static final String IOC_HISTORY_INDEX_PATTERN = "<." + IOC_INDEX_NAME_BASE + "-history-{now/d{yyyy.MM.dd.hh.mm.ss|UTC}}-1>";
+
     private CorrelationRuleIndices correlationRuleIndices;
```
eirsep, line 125: have to register batch size setting. let's give it a generic name. we can use it for all bulk requests

src/main/java/org/opensearch/securityanalytics/model/IocDao.java
```diff
@@ -172,6 +173,11 @@ public static IocDao parse(XContentParser xcp, String id) throws IOException {
             xcp.nextToken();
 
             switch (fieldName) {
+                case ID_FIELD:
+                    if (Objects.equals(id, NO_ID)) {
```
eirsep, line 177: why this check?

src/main/java/org/opensearch/securityanalytics/model/IocDao.java
```diff
@@ -172,6 +173,11 @@ public static IocDao parse(XContentParser xcp, String id) throws IOException {
             xcp.nextToken();
 
             switch (fieldName) {
+                case ID_FIELD:
```
eirsep, line 176: why are we allowing id to be empty or null at all?

src/main/java/org/opensearch/securityanalytics/services/IocService.java
```diff
@@ -0,0 +1,131 @@
...
+    public void indexIocs(List<IocDao> allIocs, ActionListener<FetchIocsActionResponse> listener) throws IOException {
+        // TODO hurneyt this is using TIF batch size setting. Consider adding IOC-specific setting
+        Integer batchSize = this.clusterService.getClusterSettings().get(SecurityAnalyticsSettings.BATCH_SIZE);
+
+        List<BulkRequest> bulkRequestList = new ArrayList<>();
+        BulkRequest bulkRequest = new BulkRequest();
+        bulkRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+
+        for (IocDao ioc : allIocs) {
+            initSystemIndexes(ioc.getType().getSystemIndexName(), listener);
```
eirsep, line 92: why is init call inside for loop

src/main/java/org/opensearch/securityanalytics/services/IocService.java
```diff
@@ -0,0 +1,131 @@
...
+     * @return TRUE if the index is an IOC-related system index, and exists; else returns FALSE.
+     */
+    public boolean hasIocSystemIndex(String index) {
+        return index.startsWith(IOC_INDEX_NAME_BASE) && this.clusterService.state().routingTable().hasIndex(index);
+    }
+
+    public void initSystemIndexes(String index, ActionListener<FetchIocsActionResponse> listener) {
+        if (!hasIocSystemIndex(index)) {
+            var indexRequest = new CreateIndexRequest(index)
+                    // TODO hurneyt finalize mappings once IOC data model PR is merged
```
eirsep, line 64: PR is merged. you can add a mapping and we can iteratively make any edits to it as before we merge feature branch. let's get it in so we don't miss out

src/main/java/org/opensearch/securityanalytics/services/IocService.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    public void indexIocs(List<IocDao> allIocs, ActionListener<FetchIocsActionResponse> listener) throws IOException {
+        // TODO hurneyt this is using TIF batch size setting. Consider adding IOC-specific setting
+        Integer batchSize = this.clusterService.getClusterSettings().get(SecurityAnalyticsSettings.BATCH_SIZE);
+
+        List<BulkRequest> bulkRequestList = new ArrayList<>();
+        BulkRequest bulkRequest = new BulkRequest();
+        bulkRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
```
eirsep, line 89: NIT: since we are batching bulk requests 
can you refresh only once after all bulk requests are submitted or on the last bulk request only

eirsep (changes requested):

src/main/java/org/opensearch/securityanalytics/services/IocService.java
eirsep, line 92: `ioc.getType().getSystemIndexName()`  -> this logic to derive system index name doesn't seem right

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
eirsep: true but the intersection of all feeds might give us some Ioc's which we can consider high serverity Iocs
the duplication would help us better with Ioc Insights and threat confidence of Iocs - number of feeds marking ioc as malicious

AWSHurneyt: No problem. Just wanted to confirm that was the intention.

@eirsep Also to clarify, by `feed id`, you're referring to the ID of the sourceConfig, correct? E.g., https://github.com/opensearch-project/security-analytics/blob/feature/threat_intel/src/main/java/org/opensearch/securityanalytics/threatIntel/model/SATIFSourceConfig.java#L67

build.gradle
```diff
@@ -21,6 +21,7 @@ buildscript {
             opensearch_build += "-SNAPSHOT"
         }
         common_utils_version = System.getProperty("common_utils.version", opensearch_build)
+        sa_commons_version = System.getProperty("sa_commons.version", opensearch_build)
```
eirsep, line 24: Will not be opensearch based

eirsep (changes requested):

build.gradle
AWSHurneyt: Removed the build.gradle changes for now. Will add them in a separate PR.

src/main/java/org/opensearch/securityanalytics/services/IocService.java
AWSHurneyt: This will be removed. I added this function when I was thinking of creating an index for each IOC type.

src/main/java/org/opensearch/securityanalytics/model/IocDao.java
AWSHurneyt: Will revise.

src/main/java/org/opensearch/securityanalytics/services/IocService.java
AWSHurneyt: The init function performs a check to see if the index exists, and then only creates the index if it doesn't. I initially performed the init check for each IOC because I was planning to store IOCs in type-specific indexes.

Since we're instead planning to store IOCs in feedSourceConfig-specific indexes; I've move the init call outside of the loop.

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
AWSHurneyt: > No problem. Just wanted to confirm that was the intention.
> 
> @eirsep Also to clarify, by `feed id`, you're referring to the ID of the sourceConfig, correct? E.g., https://github.com/opensearch-project/security-analytics/blob/feature/threat_intel/src/main/java/org/opensearch/securityanalytics/threatIntel/model/SATIFSourceConfig.java#L67

Confirmed offline that `feed id` refers to the sourceConfig ID.

src/main/java/org/opensearch/securityanalytics/model/IocDao.java
AWSHurneyt: My initial thought was that we needed to read the ID from the xcp; but I don't think that's needed anymore. Will remove.

src/main/java/org/opensearch/securityanalytics/model/IocDao.java
AWSHurneyt: My initial thought was that we needed to read the ID from the xcp; but I don't think that's needed anymore. Will remove.

src/main/java/org/opensearch/securityanalytics/services/IocService.java
AWSHurneyt: @eirsep I added the `startsWith` check just to confirm that the function is only being used to create indexes for the purposes of this feature. Do you think that check should be removed?

I implemented it as public just out of habit in case I might need it elsewhere; I'll change it to private.

src/main/java/org/opensearch/securityanalytics/services/IocService.java
AWSHurneyt: @eirsep The `FetchIocsActionResponse` is an artifact from when I was initially implementing this as a rest API. After discussing it with @jowg-amazon, she'll be passing the service an ActionListener, so I'll refactor this to a generic `ActionResponse` object for now and leave a TODO.

src/main/java/org/opensearch/securityanalytics/services/IocService.java
AWSHurneyt: Revised.

src/main/java/org/opensearch/securityanalytics/services/IocService.java
AWSHurneyt: Yes. Will separate the functions into another class in a separate PR.

src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
AWSHurneyt: Clarified offline that using the existing batch setting on this line is sufficient for now.
https://github.com/opensearch-project/security-analytics/pull/1048/files#diff-9aa7b128ff9bc2c4d9f95581a20d89a90be23c1b9b9a4f29b7271daa5f826cd6R105

src/main/resources/mappings/ioc_mapping.json
```diff
@@ -0,0 +1,54 @@
...
+  "properties": {
+    "ioc": {
+      "type": "nested",
+      "dynamic": "false",
+      "properties": {
+        "id": {
+          "type": "keyword"
+        },
+        "name": {
+          "type": "text",
```
eirsep, line 14: why cant name be keyword like id?

src/main/resources/mappings/ioc_mapping.json
```diff
@@ -0,0 +1,54 @@
...
+        },
+        "severity": {
+          "type": "keyword"
+        },
+        "specVersion": {
+          "type": "keyword"
+        },
+        "created": {
+          "type": "date",
+          "format": "strict_date_optional_time||epoch_millis"
```
eirsep, line 36: do we need this restriction?

src/main/resources/mappings/ioc_mapping.json
```diff
@@ -0,0 +1,54 @@
...
+        },
+        "modified": {
+          "type": "date",
+          "format": "strict_date_optional_time||epoch_millis"
```
eirsep, line 40: do we need this restriction?

eirsep (changes requested):
can you check some minor comments?

src/main/java/org/opensearch/securityanalytics/action/FetchIocsActionResponse.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+
+public class FetchIocsActionResponse extends ActionResponse implements ToXContentObject {
+    public static String IOCS_FIELD = "iocs";
+    public static String TOTAL_FIELD = "total";
+    private List<IocDto> iocs = Collections.emptyList();
+
+    public FetchIocsActionResponse(List<IOC> iocs) {
+        super();
+        iocs.forEach( ioc -> this.iocs.add(new IocDto(ioc)));
```
jowg-amazon, line 27: Should this be done in transport/service class?

AWSHurneyt: @jowg-amazon It can be. The `FetchIocsActionResponse` object is just a placeholder I've been using to test the changes. We can refactor and remove it before launch.

src/main/resources/mappings/ioc_mapping.json
AWSHurneyt: Revised to keyword type.

src/main/resources/mappings/ioc_mapping.json
AWSHurneyt: We use similar format for the mappings of some other plugin assets (e.g., [detectors](https://github.com/opensearch-project/security-analytics/blob/main/src/main/resources/mappings/detectors.json#L70)). Did you have another format in mind?

For now, I've revised this to `strict_date_time||epoch_millis`. My thoughts are that the create date should be required whereas the modified date could be null if the entry hasn't been updated.

src/main/resources/mappings/ioc_mapping.json
AWSHurneyt: We use similar format for the mappings of some other plugin assets (e.g., [detectors](https://github.com/opensearch-project/security-analytics/blob/main/src/main/resources/mappings/detectors.json#L70)). Did you have another format in mind?

AWSHurneyt:
> can you check some minor comments?

@eirsep Responded.

src/main/java/org/opensearch/securityanalytics/action/FetchIocsActionResponse.java
```diff
@@ -0,0 +1,50 @@
...
+import org.opensearch.core.xcontent.ToXContentObject;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.securityanalytics.model.IOC;
+import org.opensearch.securityanalytics.model.IocDto;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+
+public class FetchIocsActionResponse extends ActionResponse implements ToXContentObject {
```
eirsep, line 20: NIT: let's keep consistent naming: ListIocsResponse?

src/main/java/org/opensearch/securityanalytics/model/IOC.java
```diff
@@ -22,12 +22,8 @@
 import java.util.List;
 import java.util.Locale;
 
-import static org.opensearch.securityanalytics.SecurityAnalyticsPlugin.IOC_DOMAIN_INDEX_NAME;
-import static org.opensearch.securityanalytics.SecurityAnalyticsPlugin.IOC_HASH_INDEX_NAME;
-import static org.opensearch.securityanalytics.SecurityAnalyticsPlugin.IOC_IP_INDEX_NAME;
-
-public class IocDao implements Writeable, ToXContentObject {
-    private static final Logger logger = LogManager.getLogger(IocDao.class);
+public class IOC implements Writeable, ToXContentObject {
```
eirsep, line 25: how come Ioc is a concrete class and also STIX2Ioc?

src/main/java/org/opensearch/securityanalytics/model/IocDto.java
```diff
@@ -76,22 +73,22 @@ public void writeTo(StreamOutput out) throws IOException {
...
+                .field(IOC.SPEC_VERSION_FIELD, specVersion)
+                .timeField(IOC.CREATED_FIELD, created)
+                .timeField(IOC.MODIFIED_FIELD, modified)
+                .field(IOC.DESCRIPTION_FIELD, description)
+                .field(IOC.LABELS_FIELD, labels)
+                .field(IOC.FEED_ID_FIELD, feedId)
                 .endObject();
     }
 
     public static IocDto parse(XContentParser xcp, String id) throws IOException {
```
eirsep, line 90: this defeats the purpose of having separate classes for DTO and the model.. if there is ever a divergence we would end up having to re-write this method

eirsep (commented):
Plz invoke IocService from TIFConfigService

PR #1077, opened 2024-06-14
Assets for IOC fetch and index service.

AWSHurneyt:
### Description
Implemented assets for IOC fetch and index service.
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1078, opened 2024-06-14
Refresh API and adds Update TIF Source Config logic

jowg-amazon:
### Description
This PR adds the refresh API and refactors some logic from the source config runner to the service class. It also adds logic to update the source config.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/securityanalytics/threatIntel/action/SARefreshTIFSourceConfigAction.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.opensearch.action.ActionType;
+import org.opensearch.action.support.master.AcknowledgedResponse;
+
+import static org.opensearch.securityanalytics.threatIntel.sacommons.IndexTIFSourceConfigAction.REFRESH_TIF_SOURCE_CONFIG_ACTION_NAME;
+
+/**
+ * Refresh TIF Source Config Action
+ */
+public class SARefreshTIFSourceConfigAction extends ActionType<AcknowledgedResponse> {
```
eirsep, line 16: just `RefreshTIFSourceConfigAction` ?

src/main/java/org/opensearch/securityanalytics/threatIntel/action/SARefreshTIFSourceConfigAction.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.securityanalytics.threatIntel.action;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.action.support.master.AcknowledgedResponse;
```
eirsep, line 9: plz use  org.opensearch.extensions.AcknowledgedResponse

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFSourceConfigRunner.java
```diff
@@ -121,53 +122,12 @@ protected Runnable retrieveLockAndUpdateConfig(final SATIFSourceConfig SaTifSour
...
-                                        }, ex -> {
-                                            log.error("Failed to update threat intel source config [{}]", SaTifSourceConfig.getId());
-                                            listener.onFailure(ex);
-                                        }
-                                ));
-                                log.error("Failed to download and save IOCs for threat intel source config [{}]", SaTifSourceConfig.getId());
-                                listener.onFailure(e);
-                            }
-                    ));
+        SaTifSourceConfigManagementService.refreshTIFSourceConfig(SaTifSourceConfig.getId(), ActionListener.wrap(
```
eirsep, line 125: why did we remove the existing update code and call refresh in `updateSourceConfigAndIOCs`?

src/main/java/org/opensearch/securityanalytics/threatIntel/resthandler/RestRefreshTIFSourceConfigAction.java
```diff
@@ -0,0 +1,51 @@
...
+    private static final Logger log = LogManager.getLogger(RestRefreshTIFSourceConfigAction.class);
+
+    @Override
+    public String getName() {
+        return "refresh_tif_config_action";
+    }
+
+    @Override
+    public List<Route> routes() {
+        return List.of(new Route(RestRequest.Method.POST, String.format(Locale.getDefault(), "%s/refresh/{%s}", SecurityAnalyticsPlugin.THREAT_INTEL_SOURCE_URI, THREAT_INTEL_SOURCE_CONFIG_ID)));
```
eirsep, line 32: should be `"%s/%s/refresh"`

and why is there a surrounding curly braces on the 2nd %s placeholder??

eirsep (changes requested):

src/main/java/org/opensearch/securityanalytics/threatIntel/resthandler/RestRefreshTIFSourceConfigAction.java
AWSHurneyt: Should the route include `_` (e.g., `%s/{%s}/_refresh`), similar to the execute monitor API?
https://opensearch.org/docs/latest/observing-your-data/alerting/api/#run-monitor

src/main/java/org/opensearch/securityanalytics/threatIntel/action/SARefreshTIFSourceConfigAction.java
jowg-amazon: There should be an implementation of this in sa commons that `SARefreshTIFSourceConfigAction` will extend from but will write all the interfaces at once in a separate PR

src/main/java/org/opensearch/securityanalytics/threatIntel/action/SARefreshTIFSourceConfigAction.java
jowg-amazon: `org.opensearch.extensions.AcknowledgedResponse` is not allowed as the abstract class type, we currently use `org.opensearch.action.support.master.AcknowledgedResponse` everywhere in the repo when returning just the actionResponse https://github.com/opensearch-project/security-analytics/blob/362f0d6c1283756ae21ff9858bcf88daf938d67e/src/main/java/org/opensearch/securityanalytics/action/DeleteCorrelationRuleAction.java#L12
I can add a response class that returns different information (like config id etc...) if `org.opensearch.extensions.AcknowledgedResponse` shouldn't be used

src/main/java/org/opensearch/securityanalytics/threatIntel/resthandler/RestRefreshTIFSourceConfigAction.java
jowg-amazon: We seem to add the curly braces around any ID field throughout the repo so did it for consistency. Will change the ordering so that it's `%s/{%s}/_refresh`

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFSourceConfigRunner.java
jowg-amazon: Originally refactored this becaues the logic for the refresh api and the job scheduler were the same but if we're adding a background job to delete unused indices realizing the logic might be slightly different now. Changing it back to existing code but may refactor again later to see what logic can be reused

eirsep (changes requested):
For Create tif source config API:
we need to 
1. create tif source config and index it with state as CREATING
2. invoke ioc download.
3. update as AVAILABLE if download succeeds. else delete feed from index and return ERROR.

AWSHurneyt:
> For Create tif source config API: we need to
> 
> 1. create tif source config and index it with state as CREATING
> 2. invoke ioc download.
> 3. update as AVAILABLE if download succeeds. else delete feed from index and return ERROR.

This is in reference to https://github.com/opensearch-project/security-analytics/pull/1077/files#diff-e2af2dcb9a2eff5b384869ea9c85d1649dd3b63ad03ebe2963c1f304dec40402R114

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFSourceConfigRunner.java
```diff
@@ -122,47 +124,47 @@ protected Runnable retrieveLockAndUpdateConfig(final SATIFSourceConfig SaTifSour
...
+                saTifSourceConfigResponse -> {
+                    if (saTifSourceConfigResponse == null) {
+                        log.info("Threat intel source config [{}] does not exist", SaTifSourceConfig.getId());
                         return;
                     }
-                    if (TIFJobState.AVAILABLE.equals(SaTifSourceConfigResponse.getState()) == false) {
-                        log.error("Invalid TIF job state. Expecting {} but received {}", TIFJobState.AVAILABLE, SaTifSourceConfigResponse.getState());
+
+                    if (TIFJobState.AVAILABLE.equals(saTifSourceConfigResponse.getState()) == false) {
+                        log.error("Invalid TIF job state. Expecting {} but received {}", TIFJobState.AVAILABLE, saTifSourceConfigResponse.getState());
```
eirsep, line 134: what if state is not "AVAILABLE>

src/main/java/org/opensearch/securityanalytics/threatIntel/action/SARefreshTIFSourceConfigRequest.java
```diff
@@ -0,0 +1,61 @@
...
+
+import static org.opensearch.action.ValidateActions.addValidationError;
+import static org.opensearch.securityanalytics.threatIntel.common.Constants.THREAT_INTEL_SOURCE_CONFIG_ID;
+
+/**
+ * Refresh threat intel feed source config request
+ */
+public class SARefreshTIFSourceConfigRequest extends ActionRequest {
+    private final String id;
+    private final Long version;
```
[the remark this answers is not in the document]
jowg-amazon: taking version out for now as it's not being used in the refresh flow

src/main/java/org/opensearch/securityanalytics/threatIntel/jobscheduler/TIFSourceConfigRunner.java
jowg-amazon: If state isn't available, then it could mean that it's either currently refreshing, deleting, retrying (state will be added in later PR), or that something has failed (create_failed, refresh_failed). In these scenarios we just want to skip the current run of the job scheduler

jowg-amazon:
@eirsep Updated the logic when creating the source config so that we index the CREATING state first before downloading and saving the IOCS, added a TODO to delete ioc indices if failed

src/main/java/org/opensearch/securityanalytics/threatIntel/action/SARefreshTIFSourceConfigAction.java
eirsep: nvm lets go with this

src/main/java/org/opensearch/securityanalytics/threatIntel/action/SARefreshTIFSourceConfigAction.java
eirsep: actions won't have implementations.
just derive action names from sa commons

AWSHurneyt (approved):

PR #1087, opened 2024-06-20
Fix ioc store config mappings

jowg-amazon:
### Description
[Describe what this change achieves]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/services/STIX2IOCFeedStore.java
+++ b/src/main/java/org/opensearch/securityanalytics/services/STIX2IOCFeedStore.java
@@ -112,14 +112,17 @@ public class STIX2IOCFeedStore implements FeedStore {
     }
 
     public void indexIocs(List<STIX2IOC> iocs) throws IOException {
-        String feedIndexName = initFeedIndex(saTifSourceConfig.getId());
+        String feedIndexName = getFeedConfigIndexName(saTifSourceConfig.getId());
 
-        // Add the created index to the IocStoreConfig
-        saTifSourceConfig.getIocTypes().forEach(type -> {
-            String lowerCaseType = type.toLowerCase(Locale.ROOT);
-            ((DefaultIocStoreConfig) saTifSourceConfig.getIocStoreConfig()).getIocMapStore().putIfAbsent(lowerCaseType, new ArrayList<>());
-            ((DefaultIocStoreConfig) saTifSourceConfig.getIocStoreConfig()).getIocMapStore().get(lowerCaseType).add(feedIndexName);
-        });
+        // init index and add name to ioc map store only if index does not already exist, otherwise ioc map store will contain duplicate index names
+        if (feedIndexExists(feedIndexName) == false) {
+            initFeedIndex(feedIndexName);
+            saTifSourceConfig.getIocTypes().forEach(type -> {
+                String lowerCaseType = type.toLowerCase(Locale.ROOT);
+                ((DefaultIocStoreConfig) saTifSourceConfig.getIocStoreConfig()).getIocMapStore().putIfAbsent(lowerCaseType, new ArrayList<>());
+                ((DefaultIocStoreConfig) saTifSourceConfig.getIocStoreConfig()).getIocMapStore().get(lowerCaseType).add(feedIndexName);
+            });
+        }
 
         List<BulkRequest> bulkRequestList = new ArrayList<>();
         BulkRequest bulkRequest = new BulkRequest();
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/securityanalytics/services/STIX2IOCFeedStore.java
```diff
@@ -115,8 +115,11 @@ public void indexIocs(List<STIX2IOC> iocs) throws IOException {
         String feedIndexName = initFeedIndex(saTifSourceConfig.getId());
 
         // Add the created index to the IocStoreConfig
-        ((DefaultIocStoreConfig) saTifSourceConfig.getIocStoreConfig()).getIocMapStore().putIfAbsent(saTifSourceConfig.getId(), new ArrayList<>());
-        ((DefaultIocStoreConfig) saTifSourceConfig.getIocStoreConfig()).getIocMapStore().get(saTifSourceConfig.getId()).add(feedIndexName);
+        saTifSourceConfig.getIocTypes().forEach(type -> {
```
AWSHurneyt, line 118: Not blocking, but we should probably considering refactoring the `IocTypes` list in the config data model to store a list of the IOCType enums instead of Strings.
https://github.com/opensearch-project/security-analytics-commons/blob/main/tif/src/main/java/org/opensearch/securityanalytics/commons/model/IOCType.java#L10

src/main/java/org/opensearch/securityanalytics/threatIntel/model/DefaultIocStoreConfig.java
```diff
@@ -25,7 +25,7 @@ public class DefaultIocStoreConfig extends IocStoreConfig implements Writeable,
     public static final String DEFAULT_FIELD = "default";
     public static final String IOC_MAP = "ioc_map";
 
-    // Maps the SATIFSourceConfig ID to the list of index/alias names
+    // Maps the IOCs to the list of index/alias names
```
AWSHurneyt, line 28: Would this be "IOC types" to the index/alias names?

jowg-amazon: yes fixing comment to include ioc types

src/test/java/org/opensearch/securityanalytics/resthandler/SATIFSourceConfigRestApiIT.java
```diff
@@ -159,16 +159,16 @@ public void testCreateSATIFSourceConfigAndVerifyJobRan() throws IOException, Int
...
-            } catch (IOException e) {
-                throw new RuntimeException("failed to verify that job ran");
-            }
-        }, 240, TimeUnit.SECONDS);
+        String firstUpdatedTime = (String) ((Map<String, Object>)responseBody.get("source_config")).get("last_update_time");
+
+//        // wait for job runner to run
+//        waitUntil(() -> {
+//            try {
+//                return verifyJobRan(createdId, firstUpdatedTime);
```
AWSHurneyt, line 167: Not blocking, but could you add a javadoc comment to the `verifyJobRan` function to help clarify its use?

AWSHurneyt (approved):

src/test/java/org/opensearch/securityanalytics/resthandler/SATIFSourceConfigRestApiIT.java
jowg-amazon: added

eirsep (approved):

src/main/java/org/opensearch/securityanalytics/services/STIX2IOCFeedStore.java
AWSHurneyt, line 118: @jowg-amazon A quick way to resolve the bug where the index name is being added each time the feed is refreshed would be to add the following add this following check around the `forEach` loop.
```
if (feedIndexExists(feedIndexName)) {
...
}

jowg-amazon: Would this always be true since we are doing `initFeedIndex` first?

AWSHurneyt: Actually, it would be better to refactor the `initFeedIndex` function to take in the index name, and a boolean "exists" variable; and replace the call to `feedIndexExists` in the init function with the boolean variable.

jowg-amazon: Refactored method so that it checks if the index exists first, if it does not exist then it will initialize the index and add it to the ioc map

PR #1173, opened 2024-07-18
[BUG] Resolve aliases in monitor input to concrete indices before computing ioc-containing fields from concrete index docs

eirsep:
BUG: threat intel monitor not working on aliases in input

FIX: Currently `per_ioc_type_scan_input_list` field in monitor contains a map of index to the fields. This index input could be an alias. while executing monitor and dealing with concrete indices we would have to re-resolve the indices to create mapping of conrete index to the monitor input indices


sbcd90 (dismissed):

AWSHurneyt (approved):
Approves assuming CI passes

jowg-amazon (approved):

PR #1179, opened 2024-07-18
Test PR to run security integ tests

cwperks:
### Description

Testing a small change to see how it affects security CI checks
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
+++ b/src/main/java/org/opensearch/securityanalytics/SecurityAnalyticsPlugin.java
@@ -35,7 +35,6 @@ import org.opensearch.index.IndexSettings;
 import org.opensearch.index.codec.CodecServiceFactory;
 import org.opensearch.index.engine.EngineFactory;
 import org.opensearch.index.mapper.Mapper;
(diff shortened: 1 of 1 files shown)


cwperks:
FYI The security plugin's main branch runs targets JDK 21 only. For the main branch you can remove 11 and 17 from the matrix

cwperks:
Looks like this fixed the CI check.

Make sure that your plugin wraps all instances of interactions with a system index in:

```
try (StoredContext ctx = threadContext.stashContext()) {
    // Interact with system index here
}
```

This [change](https://github.com/opensearch-project/security/issues/4535) is going into 2.16 and it looks like it broke the tests here.

The reason why this was working <= 2.15 is because security was only considering this [list](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java#L47-L78) of system indices and the SAP Threat intel index was not tracked in that list.

From 2.16 onwards, all system indices registered via SystemIndexPlugin.getSystemIndexDescriptor will also get system index protection from the security plugin by default.

Issue #1255, opened 2024-08-16
[BUG] Empty IOC indices created even when source config index creation fails


jowg-amazon:
**What is the bug?**
When creating a threat intel source config, if downloading ioc indices is not successful, the ioc index still gets created as an empty index even though the source config is deleted.

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - Version: 2.15
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

PR #1267, opened 2024-08-21
Clean up empty IOC indices created by failed source configs

jowg-amazon:
### Description
Deletes and cleans up any ioc indices that were created by failed source configs. Also manually deletes the job scheduler lock for source configs that have the job disabled. When a source config job is enabled and then the source config is deleted, the lock is automatically deleted. However, the lock is not cleaned up if the job is disabled. This PR adds a change to check whether or not the job is enabled, if not then it will manually delete the associated lock when source config is deleted.

### Related Issues
Resolves https://github.com/opensearch-project/security-analytics/issues/1255

### Check List

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java
+++ b/src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java
@@ -389,7 +389,7 @@ public class SATIFSourceConfigService {
 
         // check to make sure the job scheduler lock index exists
         if (clusterService.state().metadata().hasIndex(LOCK_INDEX_NAME) == false) {
-            actionListener.onFailure(SecurityAnalyticsException.wrap(new OpenSearchStatusException("Threat intel job scheduler lock index does not exist", RestStatus.BAD_REQUEST)));
+            actionListener.onResponse(null);
             return;
         }
 
@@ -404,7 +404,8 @@ public class SATIFSourceConfigService {
                         log.info("Deleted threat intel job scheduler lock [{}] successfully", id);
                         actionListener.onResponse(deleteResponse);
                     } else if (deleteResponse.status().equals(RestStatus.NOT_FOUND)) {
-                        actionListener.onFailure(SecurityAnalyticsException.wrap(new OpenSearchStatusException(String.format(Locale.getDefault(), "Threat intel job scheduler lock with id [{%s}] not found", id), RestStatus.NOT_FOUND)));
+                        log.info("Threat intel job scheduler lock with id [{}] not found", id);
+                        actionListener.onResponse(deleteResponse);
                     } else {
                         actionListener.onFailure(SecurityAnalyticsException.wrap(new OpenSearchStatusException(String.format(Locale.getDefault(), "Failed to delete threat intel job scheduler lock with id [{%s}]", id), deleteResponse.status())));
                     }
--- a/src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java
+++ b/src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java
@@ -72,6 +72,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static org.opensearch.jobscheduler.spi.utils.LockService.LOCK_INDEX_NAME;
 import static org.opensearch.securityanalytics.settings.SecurityAnalyticsSettings.INDEX_TIMEOUT;
 import static org.opensearch.securityanalytics.threatIntel.common.TIFJobState.AVAILABLE;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java
```diff
@@ -376,6 +377,44 @@ public void deleteTIFSourceConfig(
...
+            return;
+        }
+
+        // check to make sure the job scheduler lock index exists
+        if (clusterService.state().metadata().hasIndex(LOCK_INDEX_NAME) == false) {
+            actionListener.onFailure(SecurityAnalyticsException.wrap(new OpenSearchStatusException("Threat intel job scheduler lock index does not exist", RestStatus.BAD_REQUEST)));
+            return;
+        }
+
+        String id = SecurityAnalyticsPlugin.JOB_INDEX_NAME + "-" + saTifSourceConfig.getId();
```
AWSHurneyt, line 396: Just to clarify, we don't need to convert this string to lowercase because this ID of the document in the `JOB_INDEX_NAME` is literally `SecurityAnalyticsPlugin.JOB_INDEX_NAME + "-" + saTifSourceConfig.getId()`, right?

jowg-amazon: Yes, here's an example of what the id in the job scheduler lock index looks like `.opensearch-sap--job-Q3l1dpEBFBLkuvfJmI5a`

AWSHurneyt (approved):

src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java
```diff
@@ -376,6 +377,44 @@ public void deleteTIFSourceConfig(
...
+            final ActionListener<DeleteResponse> actionListener
+    ) {
+        if (saTifSourceConfig.isEnabled()) {
+            actionListener.onResponse(null);
+            return;
+        }
+
+        // check to make sure the job scheduler lock index exists
+        if (clusterService.state().metadata().hasIndex(LOCK_INDEX_NAME) == false) {
+            actionListener.onFailure(SecurityAnalyticsException.wrap(new OpenSearchStatusException("Threat intel job scheduler lock index does not exist", RestStatus.BAD_REQUEST)));
```
engechas, line 392: I've seen this done both ways and I think either can make sense based on the situation. We can either

1. Treat the index missing as a failure since we intended to delete something and could not (current approach)
2. Treat the index missing as success since we intended to delete something but it did not exist so we have fulfilled the request to delete it

It would be worth considering when this API would be invoked to decide whether 1 or 2 is the correct behavior. I could see things like operator intervention resulting in the lock index being deleted and we should be sure that would not cause permanent failures/further issues if we go with approach 1.

src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java
```diff
@@ -376,6 +377,44 @@ public void deleteTIFSourceConfig(
...
+                        actionListener.onResponse(deleteResponse);
+                    } else if (deleteResponse.status().equals(RestStatus.NOT_FOUND)) {
+                        actionListener.onFailure(SecurityAnalyticsException.wrap(new OpenSearchStatusException(String.format(Locale.getDefault(), "Threat intel job scheduler lock with id [{%s}] not found", id), RestStatus.NOT_FOUND)));
+                    } else {
+                        actionListener.onFailure(SecurityAnalyticsException.wrap(new OpenSearchStatusException(String.format(Locale.getDefault(), "Failed to delete threat intel job scheduler lock with id [{%s}]", id), deleteResponse.status())));
+                    }
+                }, e -> {
+                    log.error("Failed to delete threat intel job scheduler lock with id [{}]", id);
+                    actionListener.onFailure(e);
+                }
```
engechas, line 414: Let's make sure we have unit tests for all of these scenarios

engechas (dismissed):

src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java
jowg-amazon: Makes sense, this is only called when we want to clean up and delete everything related to the source config that's being deleted so i think the second option makes more sense in this case. Updated the code to not fail when the index doesn't exist or if the document is not found.

src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java
jowg-amazon: Yep, will do in a follow up PR and added a github issue to increase our code cov and integ testing overall: https://github.com/opensearch-project/security-analytics/issues/1268

engechas (approved):

AWSHurneyt (approved):

PR #1324, opened 2024-09-25
separate doc-level monitor query indices created by detectors

sbcd90:
### Description
separate doc-level monitor query indices created by detectors

### Related Issues

### Check List
- [X] Commits are signed per the DCO using `--signoff`.


src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -887,7 +887,7 @@ private IndexMonitorRequest createDocLevelMonitorMatchAllRequest(
 
         Monitor monitor = new Monitor(monitorId, Monitor.NO_VERSION, monitorName, false, detector.getSchedule(), detector.getLastUpdateTime(), null,
                 Monitor.MonitorType.DOC_LEVEL_MONITOR.getValue(), detector.getUser(), 1, docLevelMonitorInputs, triggers, Map.of(),
-                new DataSources(detector.getRuleIndex(),
+                new DataSources(detector.getRuleIndex() + "_chained_findings",
```
eirsep, line 890: why do we need separate index?

src/main/java/org/opensearch/securityanalytics/config/monitors/DetectorMonitorConfig.java
```diff
@@ -22,7 +24,11 @@ public class DetectorMonitorConfig {
     public static final String OPENSEARCH_SAP_RULE_INDEX_TEMPLATE = ".opensearch-sap-detectors-queries-index-template";
 
     public static String getRuleIndex(String logType) {
-        return String.format(Locale.getDefault(), ".opensearch-sap-%s-detectors-queries", logType);
+        return String.format(Locale.getDefault(), ".opensearch-sap-%s-detectors-queries*", logType);
+    }
+
+    public static String getRuleIndexOptimized(String logType) {
+        return String.format(Locale.getDefault(), ".opensearch-sap-%s-detectors-queries-optimized-%s", logType, UUID.randomUUID());
```
eirsep, line 31: we are moving from index patttern to index?

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -1369,7 +1369,11 @@ void onGetResponse(Detector currentDetector, User user) {
             request.getDetector().setAlertsHistoryIndexPattern(DetectorMonitorConfig.getAlertsHistoryIndexPattern(ruleTopic));
             request.getDetector().setFindingsIndex(DetectorMonitorConfig.getFindingsIndex(ruleTopic));
             request.getDetector().setFindingsIndexPattern(DetectorMonitorConfig.getFindingsIndexPattern(ruleTopic));
-            request.getDetector().setRuleIndex(DetectorMonitorConfig.getRuleIndex(ruleTopic));
+            if (currentDetector.getRuleIndex().contains("optimized")) {
```
eirsep, line 1372: if name contains `optimized` shouldn't we set rule index name as `getRuleIndexOptimized` value?

is this if else condition flipped?

src/test/java/org/opensearch/securityanalytics/DetectorThreatIntelIT.java
```diff
@@ -69,22 +69,11 @@ public void testCreateDetectorWithThreatIntelEnabled_updateDetectorWithThreatInt
         Detector detector = randomDetectorWithInputsAndThreatIntelAndTriggers(List.of(input), true, List.of(trigger));
         Response createResponse = makeRequest(client(), "POST", SecurityAnalyticsPlugin.DETECTOR_BASE_URI, Collections.emptyMap(), toHttpEntity(detector));
 
-        String request = "{\n" +
-                "   \"query\" : {\n" +
-                "     \"match_all\":{\n" +
-                "     }\n" +
-                "   }\n" +
-                "}";
-        SearchResponse response = executeSearchAndGetResponse(DetectorMonitorConfig.getRuleIndex(randomDetectorType()), request, true);
```
eirsep, line 78: why has this been removed

eirsep (changes requested):
what happens when someone updates a detector that doesnt have exclusive query index?

src/test/java/org/opensearch/securityanalytics/alerts/AlertsIT.java
```diff
@@ -825,21 +825,10 @@ public void testMultipleAggregationAndDocRules_alertSuccess() throws IOException
 
         Response createResponse = makeRequest(client(), "POST", SecurityAnalyticsPlugin.DETECTOR_BASE_URI, Collections.emptyMap(), toHttpEntity(detector));
 
-
-        String request = "{\n" +
```
eirsep, line 829: plz revert and fix test as necessary

src/test/java/org/opensearch/securityanalytics/findings/FindingIT.java
```diff
@@ -1209,21 +1209,11 @@ public void testCreateDetectorWithNotCondition_verifyFindingsAndNoFindings_succe
 
         Response createResponse = makeRequest(client(), "POST", SecurityAnalyticsPlugin.DETECTOR_BASE_URI, Collections.emptyMap(), toHttpEntity(detector));
 
-        String request = "{\n" +
```
eirsep, line 1212: plz revert

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
sbcd90: each doc-level monitor has its own query index now. So the `chained findings doc-level monitor` also has its own index.

src/main/java/org/opensearch/securityanalytics/config/monitors/DetectorMonitorConfig.java
sbcd90: no. this method is utilized by few integ tests & hence the change.

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
sbcd90: the `if-else condition` checks if the detector is already using a dedicated query index, then keep on using it. else, switch from shared query index to new dedicated query index as the detector is updated.

src/test/java/org/opensearch/securityanalytics/DetectorThreatIntelIT.java
sbcd90: query indices are created & deleted now. So, it is difficult to assert number of queries in query index now.

sbcd90:
> what happens when someone updates a detector that doesnt have exclusive query index?

The detector is updated to use exclusive query index.

src/test/java/org/opensearch/securityanalytics/alerts/AlertsIT.java
sbcd90: query indices are created & deleted now. So, it is difficult to assert number of queries in query index now.

src/test/java/org/opensearch/securityanalytics/findings/FindingIT.java
sbcd90: query indices are created & deleted now. So, it is difficult to assert number of queries in query index now.

sbcd90:
security tests fail as docker image not updated.

eirsep:
looks like we create aliases not indices from 'queryIndex` field in monitor datasources

eirsep:
can we add tests for scenarios

setting is **off**. create detector. verify query index name. turn **on** setting. update detector, verify query index name
setting is **on**. create detector. verify query index name. turn **off** setting. update detector, verify query index name. create new detector and verify query index(should be old style since setting is off)

src/main/java/org/opensearch/securityanalytics/config/monitors/DetectorMonitorConfig.java
```diff
@@ -22,7 +24,11 @@ public class DetectorMonitorConfig {
     public static final String OPENSEARCH_SAP_RULE_INDEX_TEMPLATE = ".opensearch-sap-detectors-queries-index-template";
 
     public static String getRuleIndex(String logType) {
-        return String.format(Locale.getDefault(), ".opensearch-sap-%s-detectors-queries", logType);
+        return String.format(Locale.getDefault(), ".opensearch-sap-%s-detectors-queries*", logType);
```
eirsep, line 27: why `*` 
this is an alias

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -1369,7 +1379,16 @@ void onGetResponse(Detector currentDetector, User user) {
...
+                request.getDetector().setRuleIndex(currentDetector.getRuleIndex());
```
eirsep, line 1383: plz add comment that if we turn OFF setting after turning on, updating detector will not change from optimized to normal and we need to re-create detector

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -1369,7 +1379,16 @@ void onGetResponse(Detector currentDetector, User user) {
...
+            } else {
+                if (enableDetectorWithDedicatedQueryIndices) {
+                    request.getDetector().setRuleIndex(DetectorMonitorConfig.getRuleIndexOptimized(ruleTopic));
+                } else {
+                    String ruleTopicIndex = DetectorMonitorConfig.getRuleIndex(ruleTopic);
+                    request.getDetector().setRuleIndex(ruleTopicIndex.substring(0, ruleTopicIndex.length() - 1));
```
eirsep, line 1389: this line is very confusing

src/test/java/org/opensearch/securityanalytics/SecurityAnalyticsRestTestCase.java
```diff
@@ -266,6 +266,7 @@ void setDebugLogLevel() throws IOException, InterruptedException {
 
 
         makeRequest(client(), "PUT", "_cluster/settings", Collections.emptyMap(), se, new BasicHeader("Content-Type", "application/json"));
+        updateClusterSetting("plugins.security_analytics.enable_detectors_with_dedicated_query_indices", "true");
```
eirsep, line 269: are there tests with setting turned off

src/test/java/org/opensearch/securityanalytics/mapper/MapperRestApiIT.java
```diff
@@ -1722,12 +1725,16 @@ public void testAzureMappings() throws IOException {
         DetectorInput input = new DetectorInput("windows detector for security analytics", List.of(indexName), List.of(),
                 getPrePackagedRules("azure").stream().map(DetectorRule::new).collect(Collectors.toList()));
         Detector detector = randomDetectorWithInputs(List.of(input), "azure");
-        createDetector(detector);
-
-        List<SearchHit> hits = executeSearch(".opensearch-sap-azure-detectors-queries-000001", matchAllSearchBody);
```
eirsep, line 1727: can you retain this test and add a new one - basically test for both setting turned. on and turned off

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
```diff
@@ -1104,21 +1094,11 @@ public void testCreateDetector_verifyWorkflowCreation_success_WithGroupByRulesIn
 
         Response createResponse = makeRequest(client(), "POST", SecurityAnalyticsPlugin.DETECTOR_BASE_URI, Collections.emptyMap(), toHttpEntity(detector));
 
-        String request = "{\n" +
```
eirsep, line 1107: we should not remove this code. we lose code coverage of tests. that's an anti-pattern. let's fix these tests

eirsep (changes requested):

sbcd90:
> can we add tests for scenarios
> 
> setting is **off**. create 2 detectors. verify query index name. turn **on** setting. update 1 detector, verify query index name. setting is **on**. create 1 detector. verify query index name. turn **off** setting. update detector, verify query index name. create new detector and verify query index(should be old style since setting is off)

added both test cases.

1st scenario: https://github.com/sbcd90/security-analytics/blob/c514a780d4b3f65ac828e3ff868c56f46f2a82b7/src/test/java/org/opensearch/securityanalytics/resthandler/DetectorRestApiIT.java#L1729
2nd scenario: https://github.com/sbcd90/security-analytics/blob/c514a780d4b3f65ac828e3ff868c56f46f2a82b7/src/test/java/org/opensearch/securityanalytics/resthandler/DetectorRestApiIT.java#L1831

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
sbcd90: fixed it.

src/test/java/org/opensearch/securityanalytics/mapper/MapperRestApiIT.java
sbcd90: added scenario for flag turned off in test cases. updated this test case to fetch queries from index pattern.

src/test/java/org/opensearch/securityanalytics/SecurityAnalyticsRestTestCase.java
sbcd90: added a test for setting turned off.

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
sbcd90: added a comment. added a test too.

src/main/java/org/opensearch/securityanalytics/config/monitors/DetectorMonitorConfig.java
sbcd90: fixed it.

src/main/java/org/opensearch/securityanalytics/transport/TransportIndexDetectorAction.java
```diff
@@ -1252,7 +1256,13 @@ void createDetector() {
             request.getDetector().setAlertsHistoryIndexPattern(DetectorMonitorConfig.getAlertsHistoryIndexPattern(ruleTopic));
             request.getDetector().setFindingsIndex(DetectorMonitorConfig.getFindingsIndex(ruleTopic));
             request.getDetector().setFindingsIndexPattern(DetectorMonitorConfig.getFindingsIndexPattern(ruleTopic));
-            request.getDetector().setRuleIndex(DetectorMonitorConfig.getRuleIndex(ruleTopic));
+
+            if (enableDetectorWithDedicatedQueryIndices) {
+                request.getDetector().setRuleIndex(DetectorMonitorConfig.getRuleIndexOptimized(ruleTopic));
+            } else {
+                String ruleTopicIndex = DetectorMonitorConfig.getRuleIndex(ruleTopic);
+                request.getDetector().setRuleIndex(ruleTopicIndex.substring(0, ruleTopicIndex.length()-1));
```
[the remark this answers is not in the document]
sbcd90: `ruleIndex` is changing to dedicated `ruleIndexOptimzied` now. Thats why `if-else condition`.

src/main/java/org/opensearch/securityanalytics/util/RuleTopicIndices.java
```diff
@@ -83,7 +83,8 @@ private void getAllRuleIndices(ActionListener<List<String>> listener) {
             listener.onResponse(
                     logTypes
                         .stream()
-                        .map(logType -> DetectorMonitorConfig.getRuleIndex(logType))
+                            // use index pattern here to define rule topic index template for all query indices which match the pattern
+                        .map(logType -> DetectorMonitorConfig.getRuleIndex(logType) + "*")
```
eirsep, line 87: shouldn't this be detector specific?
why do we have this method? `getAllRuleIndices`

sbcd90: a single index template containing specific settings for sigma rule executions are applied to all rule indices.

eirsep:
CI is failing

sbcd90:
> CI is failing

ci passed now.

eirsep (approved):

eirsep:
Security tests failed

AWSHurneyt (approved):

sbcd90:
hi @ELKHostmaster , please bounce(stop and start) OpenSearch process in every data node. This would cleanup those tasks. 
Please create detectors on a `rolling index alias with write index pointing to the current index where data is ingested`(https://opensearch.org/docs/latest/im-plugin/index-alias/#index-alias-options )

mimicbox:
Hi @sbcd90 

I am having a simmilar issue to this and have been following this. I want to try your suggestion but the part I do not get it is e-

> create detectors on a rolling index alias with write index pointing to the current index where data is ingested

How do I define this during detector creation? I use the API to create detectors. Are you talking about the index pattern you define in the inputs.detector_input.indices value? I currently use an alias that points to a datastream.

Thank you for any help at all!

sbcd90:
hi @mimicbox , if you're currently using an `index pattern` with your detector, just change it to an `index alias` instead. the term `rolling` refers to `index alias with rollovers`. 
e.g.
we have an index,

api_output-000001 -> currently points as write index for index alias api_output
now, index rolls over after sometime to ,
api_output-000002 , write index also moves if you're using an ISM policy for rollover. https://opensearch.org/docs/latest/im-plugin/ism/policies/#rollover

provide `api_output` as input to the detector. if there is no auto suggestion, manually type it into the text box.
Lastly, keep index alias name `api_output` a prefix of the index names behind it e.g `api_output-000001`, `api_output-000002`... so on, or `api_output-test-000001`, `api_output-test-000002`... so on.

ELKHostmaster:
@sbcd90   If I understand correctly, your saying point the detector to the index alias and use ISM to rollover the indices at regular intervals.

I'm already doing that. There is something related to the fan-out change introduced in 2.14 that I think has really caused the performance issues. 

Is this fixed in 2.18? Is upgrading from 2.16 a possible solution?

sbcd90:
@ELKHostmaster , are you using Sigma Aggregation Rules?

ELKHostmaster:
@sbcd90  do you mean - using any of the aggregation functions in sigma rules (https://help.sigmacomputing.com/docs/aggregate-functions) ?

No, I dont believe so but I am scanning through them now.

ELKHostmaster:
No correlation rules either. I've stripped it down to just our sigma detections as we've been trying to get to the bottom of the issue for the last month.

sbcd90:
hi @ELKHostmaster , please bounce(stop and start) OpenSearch process in every data node. This would cleanup those tasks.
please re-create(delete & create) the detectors. This will help cleanup old metadata.
I would suggest move to an odd release. e.g, 2.15 or 2.17 before doing these steps. They will have the latest fixes.

ELKHostmaster:
@sbcd90   thank you. yes, I have stopped/started every data node several times over the last month. its the only way to keep the cluster running at this point. I'm currently on 2.16

I thought this fix was added in 2.18. Are you saying it was backported to 2.17.1 ?  I didn't think that backports resulted in a new release package being built. So, I didn't think the fix would be included. 

Also, when you say "odd" release... is there something different about odd/even releases?

Thanks again for the responses!

sbcd90:
hi @ELKHostmaster , i dont think data nodes need to be stopped/started frequently. i would say take a thread dump of a data node when doc-level fanout tasks are stuck. That would help root cause the issue.

yes. i think all these fixes should be in `2.18`. But to stay up-to-date with future fixes, following an odd release helps.

ELKHostmaster:
@sbcd90  Good evening. I waited until a couple data nodes got really busy and took jstack dumps for you to look at. I also included a list of tasks running on the cluster. You can see a lot of tasks are running for 8 hours now and growing since the last reboot of the nodes. I appreciate any help you can give. Thanks!
[Uploading data-13.txt…]()

ELKHostmaster:
@sbcd90  Good morning. Could I ask a favor of you? Could you take a look at the thread dumps above and see if the issue is related to the fanout change? Our cluster is unusable right now with detectors enabled/running. Circuit breaker issues, java heap exceptions, nodes dropping out, 99% cpu on data nodes, etc. This was an upgrade from 2.13 where everything was fine. The only change was upgrading to 2.16. Same detectors/rules, etc. 

Do you think the fixes to fanout in 2.18 will help us? 

Thank You in advance!

PR #1372, opened 2024-10-22
fix for detectors with sigma aggregation rules

sbcd90:
### Description
fix for detectors with sigma aggregation rules

### Related Issues

### Check List
- [X] Commits are signed per the DCO using `--signoff`.
- [X] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/mapper/IndexTemplateManager.java
+++ b/src/main/java/org/opensearch/securityanalytics/mapper/IndexTemplateManager.java
@@ -115,13 +115,22 @@ public class IndexTemplateManager {
 
         upsertComponentTemplateStepListener.whenComplete( acknowledgedResponse -> {
 
-            // Find template which matches input index best
+            // Find template which matches input index best. starts by directly matching with input index and
+            // if not found matches with current write index.
             String templateName =
                     MetadataIndexTemplateService.findV2Template(
                             state.metadata(),
                             normalizeIndexName(indexName),
                             false
                     );
+            if (templateName == null) {
+                templateName =
+                        MetadataIndexTemplateService.findV2Template(
+                                state.metadata(),
+                                normalizeIndexName(cin),
+                                false
+                        );
+            }
 
             if (templateName == null) {
                 // If we find conflicting templates(regardless of priority) and that template was created by us,
@@ -181,8 +190,8 @@ public class IndexTemplateManager {
                 template = state.metadata().templatesV2().get(templateName);
                 if (template.composedOf().contains(componentName) == false) {
                     List<String> newComposedOf = new ArrayList<>(template.composedOf());
-                    List<String> indexPatterns = List.of(computeIndexPattern(indexName));
-                    ;
+                    List<String> indexPatterns = new ArrayList<>(template.indexPatterns());
+                    indexPatterns.add(computeIndexPattern(indexName));
                     newComposedOf.add(componentName);
 
                     try {
--- a/src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
+++ b/src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
@@ -12,6 +12,7 @@ import org.junit.Ignore;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.client.Request;
 import org.opensearch.client.Response;
+import org.opensearch.common.settings.Settings;
 import org.opensearch.commons.alerting.model.Monitor.MonitorType;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.search.SearchHit;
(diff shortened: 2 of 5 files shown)


src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
```diff
@@ -2153,6 +2157,109 @@ public void testCreateDetectorWithCloudtrailAggrRuleWithEcsFields() throws IOExc
         assertEquals(1, getFindingsBody.get("total_findings"));
     }
 
+    @SuppressWarnings("unchecked")
+    public void testCreateDetectorWithCloudtrailAggrRuleWithRolloverIndexAliases() throws IOException, InterruptedException {
+        updateClusterSetting("plugins.security_analytics.enable_detectors_with_dedicated_query_indices", "false");
```
eirsep, line 2162: why do we turn off this setting?

eirsep: plz change to true as that's the new default behaviour

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
```diff
@@ -2153,6 +2157,109 @@ public void testCreateDetectorWithCloudtrailAggrRuleWithEcsFields() throws IOExc
...
+        params.put("detector_id", createdId);
+        Response getFindingsResponse = makeRequest(client(), "GET", SecurityAnalyticsPlugin.FINDINGS_BASE_URI + "/_search", params, null);
+        Map<String, Object> getFindingsBody = entityAsMap(getFindingsResponse);
+
+        // Assert findings
+        assertNotNull(getFindingsBody);
+        assertEquals(1, getFindingsBody.get("total_findings"));
+
+        doRollover("ocsf_ct");
+        Thread.sleep(90000);
```
eirsep, line 2247: why should we do sleep?
after rollover is the template not applied to new index directly?

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
```diff
@@ -2153,6 +2157,109 @@ public void testCreateDetectorWithCloudtrailAggrRuleWithEcsFields() throws IOExc
...
+
+        indexDoc("ocsf_ct", "4", randomCloudtrailOcsfDoc());
```
eirsep, line 2249: before this should we not verify mappings of new write index?

eirsep (commented):
when is template applied? not at creation time? why is there sleep 90 s in test?

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
sbcd90: The Sigma Aggregation Rule we're using has timeframe set to `1m`(https://github.com/opensearch-project/security-analytics/blob/622d26eb5ddf3cb03f22f1cf0e0cb65a8a5c04cf/src/test/java/org/opensearch/securityanalytics/TestHelpers.java#L1162). So, the `sleep` ensures that the rule only considers docs from the rolled over new write index. If this fix is not present, the `alerting workflow run` doesn't generate the new finding. https://github.com/opensearch-project/security-analytics/blob/622d26eb5ddf3cb03f22f1cf0e0cb65a8a5c04cf/src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java#L2251

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
sbcd90: mappings of the first write index are verified with the first workflow run. https://github.com/opensearch-project/security-analytics/blob/622d26eb5ddf3cb03f22f1cf0e0cb65a8a5c04cf/src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java#L2235 .If alias mappings are not applied, findings will not be generated.

sbcd90:
> when is template applied? not at creation time? why is there sleep 90 s in test?

Yes. Template is applied at creation time with the `Create Mapping` api call.

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
sbcd90: yes. sorry. leftover from some experiments.

eirsep (approved):

src/test/java/org/opensearch/securityanalytics/resthandler/DetectorMonitorRestApiIT.java
eirsep: But i thought the bug was that AFTER rollover the template is not getting applied?

src/main/java/org/opensearch/securityanalytics/mapper/IndexTemplateManager.java
```diff
@@ -115,13 +115,22 @@ public void upsertIndexTemplateWithAliasMappings(
 
         upsertComponentTemplateStepListener.whenComplete( acknowledgedResponse -> {
 
-            // Find template which matches input index best
+            // Find template which matches input index best. starts by directly matching with input index and
```
riysaxen-amzn, line 118: let's use doc comments for these lines instead -> https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

sbcd90: will address this in a follow-up pr.

riysaxen-amzn (approved):

PR #1393, opened 2024-10-25
Add validation for threat intel source config

### LINKED ISSUE #1366: [BUG] PUT on source without "name" in request makes the source unable to be fetched/deleted/listed/updated
**What is the bug?**
Threat-intel Sources when modified using the API with PUT, without having the "name" in request body, makes the server throw 500 for every request that includes the source in the response, and the source can also not be DELETED

1. Utilize the API
2. Run POST _plugins/_security_analytics/threat_intel/sources (I utilized file_upload API) 
3. Run PUT  _plugins/_security_analytics/threat_intel/sources/<source_id> (Without "name" in request body")
4. Response Body: {"ok":false,"error":"[illegal_state_exception] Can't get text on a VALUE_NULL at 1:26"}

There should be a validation for this, and in case not, the source should still have a way to be deleted. Via the dashboards or API, and it should not break the other endpoints which return this ( Fetch All )

 - Plugins

Error 1
error 2

Crucially, the DELETE isn't working, so I cannot find a way to get rid of the faulty source via the API.

jowg-amazon:
### Description
Add source config validation so source config cannot be created without valid `name`, `type`, `format`, `source`, and `ioc type`. Additionally allows source configs with null values to still be read and deleted.

### Related Issues
Resolves #1366

### Check List

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/threatIntel/common/SourceConfigDtoValidator.java
+++ b/src/main/java/org/opensearch/securityanalytics/threatIntel/common/SourceConfigDtoValidator.java
@@ -24,7 +24,19 @@ public class SourceConfigDtoValidator {
     public List<String> validateSourceConfigDto(SATIFSourceConfigDto sourceConfigDto) {
         List<String> errorMsgs = new ArrayList<>();
 
-        if (sourceConfigDto.getIocTypes().isEmpty()) {
+        if (sourceConfigDto.getName() == null || sourceConfigDto.getName().isEmpty()) {
+            errorMsgs.add("Name must not be empty");
+        }
+
+        if (sourceConfigDto.getFormat() == null || sourceConfigDto.getFormat().isEmpty()) {
+            errorMsgs.add("Format must not be empty");
+        }
+
+        if (sourceConfigDto.getSource() == null) {
+            errorMsgs.add("Source must not be empty");
+        }
+
+        if (sourceConfigDto.getIocTypes() == null || sourceConfigDto.getIocTypes().isEmpty()) {
             errorMsgs.add("Must specify at least one IOC type");
         } else {
             for (String s: sourceConfigDto.getIocTypes()) {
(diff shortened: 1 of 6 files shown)


eirsep (approved):

eirsep (changes requested):
can we add test to verify https://github.com/opensearch-project/security-analytics/issues/1366 is solved

src/main/java/org/opensearch/securityanalytics/threatIntel/common/SourceConfigDtoValidator.java
```diff
@@ -24,7 +24,19 @@ public class SourceConfigDtoValidator {
     public List<String> validateSourceConfigDto(SATIFSourceConfigDto sourceConfigDto) {
```
AWSHurneyt, line 24: As a follow-up item, let's add validation for the refresh frequency.

src/main/java/org/opensearch/securityanalytics/threatIntel/common/SourceConfigDtoValidator.java
```diff
@@ -24,7 +24,19 @@ public class SourceConfigDtoValidator {
...
         List<String> errorMsgs = new ArrayList<>();
 
-        if (sourceConfigDto.getIocTypes().isEmpty()) {
+        if (sourceConfigDto.getName() == null || sourceConfigDto.getName().isEmpty()) {
```
AWSHurneyt, line 27: @jowg-amazon could we add length restrictions? The frontend uses this validation for names.
https://github.com/opensearch-project/security-analytics-dashboards-plugin/blob/main/public/utils/validation.ts#L7C14-L7C33

src/main/java/org/opensearch/securityanalytics/threatIntel/common/SourceConfigDtoValidator.java
```diff
@@ -24,7 +24,19 @@ public class SourceConfigDtoValidator {
...
+            errorMsgs.add("Name must not be empty");
+        }
+
+        if (sourceConfigDto.getFormat() == null || sourceConfigDto.getFormat().isEmpty()) {
```
AWSHurneyt, line 31: We may need to add length validation for this as well. Unless this gets validated somewhere downstream, the config could be created via backend API with something like 300k characters; which could disrupt functionality.

src/main/java/org/opensearch/securityanalytics/threatIntel/common/SourceConfigDtoValidator.java
jowg-amazon: added validation for names based on how frontend currently validates the name

src/main/java/org/opensearch/securityanalytics/threatIntel/common/SourceConfigDtoValidator.java
jowg-amazon: Added length validation

jowg-amazon:
> can we add test to verify #1366 is solved

Added parsing tests to make sure null fields can be parsed

AWSHurneyt (approved):
Changes LGTM. Approving assuming other's comments are addressed.

eirsep (approved):

Issue #1421, opened 2024-12-06
[FEATURE] Accepting custom json schema for uploading threat intel iocs


eirsep:
**Is your feature request related to a problem?**
Currently users can upload their own threat intelligence indicators(IoC's) in a single format defined by the plugin
The current format looks like this:
```
{"name":"test-domain-ioc-1","type":"domain-name","value":"example1.com","severity":"3","created":"2024-06-24T23:38:59.817536Z","modified":"2024-06-25T00:38:59.81754Z","description":"test ioc description","labels":["label1"],"spec_version":"spec1"}
{"name":"test-domain-ioc-2","type":"domain-name","value":"example2.com","severity":"3","created":"2024-06-24T23:38:59.817536Z","modified":"2024-06-25T00:38:59.81754Z","description":"test ioc description","labels":["label1"],"spec_version":"spec1"}
{"name":"test-domain-ioc-3","type":"domain-name","value":"example3.com","severity":"3","created":"2024-06-24T23:38:59.817536Z","modified":"2024-06-25T00:38:59.81754Z","description":"test ioc description","labels":["label1"],"spec_version":"spec1"}
```

Users who set up workflows to download threat intel from third-party security vendors or internally curated threat intelligence need to be able to upload their custom threat intelligence in different formats.

Users need a mechanism to define their own schema or format for the indicators they upload

**What solution would you like?**
Update the create/update threat intel source api to optionally accept a custom json format or schema at create time.

eirsep:
One possible solution to accept the JSON schema as part of the create threat intel source API is to add a field called **ioc_schema** and a boolean flag **use_default_schema** 
Let's consider the example:
```
ioc_schema: {
 "dataRoot" : "$.*[*]",
 "typeJsonPath": "$.type",
  "valueJsonPath": "$.ioc_value",
  "severityJsonPath": "$.severity"
}
```
this would provide flexibility in schema to accept any level of nesting and any format as long as they mention the key name for the required values - ioc type, ioc value and ioc severity

PR #1450, opened 2025-01-09
Adds support for uploading threat intelligence in Custom Format JSON

eirsep:
### Description
Currently we support users uploading threat intelligence in the fixed STIX2 based JSON format only.
This change adds the ability to provide a schema with JsonPath notation for the existing fields and uploading a json file that is then parsed to extract a list of threat intel iocs.

Changes the index Threat intel source API to accept an IoC Schema with all the fields having JsonPath notations. only ioc value and ioc type field are mandatory. Rest of the fields in the threat intel ioc list such as created date, modified date, name, id etc are optional.

### Related Issues
#1421


eirsep:
Closing in favor of https://github.com/opensearch-project/security-analytics/pull/1493

PR #1455, opened 2025-01-27
[2.x]Adds support for uploading threat intelligence in Custom Format JSON

eirsep:
### Description
Main branch CI blocked on CI failure : https://github.com/opensearch-project/security-analytics/pull/1450

Customers want to ingest threat intel iocs in their own JSON format. We need to provide ability in create threat intel source experience to accept JSON files with non-STIX2 format.
Accept [JsonPath](https://en.wikipedia.org/wiki/JSONPath) compliant schema as part of create threat intel source API

JSONPath is an industry standard query language for extracting values from Json.

Change to [Create threat intel source API](https://opensearch.org/docs/2.16/security-analytics/threat-intelligence/api/source/) 
**S3 download**
```
POST _plugins/_security_analytics/threat_intel/sources/
{
 "type": "S3_CUSTOM",
 "name": "example-ipv4-from-SAP-account",
 "format": "STIX2",
 "store_type": "OS",
 "enabled": "true",
 "schedule": {
  "interval": {
   "start_time": 1717097122,
   "period": "10",
   "unit": "DAYS"
  }
 },
 /** NEWLY ADDED FIELDS TO SUPPORT CUSTOM SCHEMA
    "ioc_schema": {
        "json_path_schema": {
            "type": {
                "json_path": "$.*[*].ioc_type",
                "is_key": false
            },
            "value": {
                "json_path": "$.*[*].ioc_value",
                "is_key": false
            }
        }
    }
 */
 "source": {
  "s3": {
   "bucket_name": "threat-intel-s3-test-bucket",
   "object_key": "alltypess3object",
   "region": "us-west-2",
   "role_arn": "arn:aws:iam::248279774929:role/threat_intel_s3_test_role"
  }
 },
 "ioc_types": [
  "domain-name",
  "ipv4-addr"
 ]
}
```
**IOC local file upload**
```
{
 "name": "test_ioc_upload",
 "format": "STIX",
 "type": "IOC_UPLOAD", 
 "description": null,
 "created_by_user": null,
/* //NEW FIELDS
 "source": {
 "custom_schema_ioc_upload": {
[...]

PR #1493, opened 2025-03-03
Adds support for uploading threat intelligence in Custom Format …

eirsep:
Cherry-picks from 2.x (#1455) to main

Closed original PR #1450 to main due to dep conflicts in CIs

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/model/STIX2IOC.java
+++ b/src/main/java/org/opensearch/securityanalytics/model/STIX2IOC.java
@@ -11,22 +11,17 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.common.io.stream.Writeable;
-import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.ToXContentObject;
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.core.xcontent.XContentParserUtils;
-import org.opensearch.securityanalytics.commons.model.IOCType;
 import org.opensearch.securityanalytics.commons.model.STIX2;
-import org.opensearch.securityanalytics.util.SecurityAnalyticsException;
 import org.opensearch.securityanalytics.util.XContentUtils;
 
 import java.io.IOException;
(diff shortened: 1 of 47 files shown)


src/main/java/org/opensearch/securityanalytics/model/STIX2IOC.java
```diff
@@ -204,26 +199,27 @@ public static STIX2IOC parse(XContentParser xcp, String id, Long version) throws
 
             switch (fieldName) {
                 case NAME_FIELD:
+                    if (xcp.currentToken() == XContentParser.Token.VALUE_NULL) {
```
AWSHurneyt, line 202: Jus curious. I noticed that the STIX2IOCDto class is using a `getString` helper function when parsing these fields. Could you clarify why we don't use something similar here?

src/main/java/org/opensearch/securityanalytics/services/STIX2IOCConnectorFactory.java
```diff
@@ -35,23 +36,47 @@ public STIX2IOCConnectorFactory(final InputCodecFactory inputCodecFactory, final
...
         final S3Client s3Client = s3ClientFactory.create(s3ConnectorConfig.getRoleArn(), s3ConnectorConfig.getRegion());
-        final InputCodec inputCodec = inputCodecFactory.create(feedConfiguration.getIocSchema().getModelClass(), feedConfiguration.getInputCodecSchema());
         return new S3Connector<>(s3ConnectorConfig, s3Client, inputCodec);
     }
 
-    public S3Connector<STIX2> createAmazonS3Connector(final FeedConfiguration feedConfiguration, List<String> clusterTuple) {
+    private InputCodec getInputCodec(FeedConfiguration feedConfiguration, SATIFSourceConfig satifSourceConfig) {
+        final InputCodec inputCodec;
+        if (satifSourceConfig != null && satifSourceConfig.getIocSchema() != null) {
+            logger.info("Parsing custom schema JSON from S3 for threat intel source [{}]", satifSourceConfig.getName());
```
AWSHurneyt, line 68: Nitpick: Would using the source config ID make more sense for this log? Or do we have validation in place to prevent customers from creating configs with the same name?

AWSHurneyt (approved):

AWSHurneyt:
Approved as PR to 2.x was reviewed and merged.

riysaxen-amzn (approved):

PR #1530, opened 2025-05-09
Switch guava deps from compileOnly to implementation

cwperks:
### Description

SAP has its guava dependencies set at compileOnly level which means they are not packaged into the SAP assembly and available at runtime. 

SAP uses these deps at runtime, however, the way this works is because of the extending relationship between SAP and JS plugin where the deps of the extended plugin are available to the extending plugin at runtime.

In https://github.com/opensearch-project/job-scheduler/pull/773, guava was removed from JS which means these deps are no longer available to SAP at runtime. I'm creating this PR to change the guava deps from compileOnly to implementation to ensure they are included in the assembly and available at runtime.

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -210,7 +210,7 @@ dependencies {
     implementation 'com.jayway.jsonpath:json-path:2.9.0'
     implementation 'net.minidev:json-smart:2.5.2'
     implementation 'net.minidev:accessors-smart:2.5.2'
-    compileOnly "com.google.guava:guava:32.1.3-jre"
+    implementation "com.google.guava:guava:32.1.3-jre"
 
     // TODO uncomment once SA commons is published to maven central
 //    api "org.opensearch:security-analytics-commons:${sa_commons_version}@jar"
--- a/src/main/java/org/opensearch/securityanalytics/util/RuleIndices.java
+++ b/src/main/java/org/opensearch/securityanalytics/util/RuleIndices.java
@@ -49,6 +49,7 @@ import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.client.Client;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.Charset;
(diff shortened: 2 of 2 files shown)


cwperks:
Build failures may be related to https://github.com/opensearch-project/OpenSearch/pull/17989

CC: @pranu2502

cwperks:
@kumargu @pranu2502 

This is something that needs to be considered for the AccessController replacement. Resources on the classpath should be exempt from permissions checks. 

In java, I believe this is handled in the ClassLoader: https://github.com/openjdk/jdk21u/blob/master/src/java.base/share/classes/jdk/internal/loader/BuiltinClassLoader.java#L500-L505

cwperks:
> @kumargu @pranu2502 @RajatGupta02
> 
> This is something that needs to be considered for the AccessController replacement. Resources on the classpath should be exempt from permissions checks.
> 
> In java, I believe this is handled in the ClassLoader: https://github.com/openjdk/jdk21u/blob/master/src/java.base/share/classes/jdk/internal/loader/BuiltinClassLoader.java#L500-L505

CC: @reta putting this on your radar too.

cwperks:
We're going to need a solution to this to support JDK 24 in this repo, because the calls to AccessController.doPrivileged within the JDK have now been removed: https://github.com/openjdk/jdk24u/commit/db7ee3dad1d9c9578794d946dd5de1f51d77e5a1

reta:
> We're going to need a solution to this to support JDK 24 in this repo, because the calls to AccessController.doPrivileged within the JDK have now been removed: [openjdk/jdk24u@db7ee3d](https://github.com/openjdk/jdk24u/commit/db7ee3dad1d9c9578794d946dd5de1f51d77e5a1)

Have you seen the failures related to that? For the limited coverage we have now, it seems like we are fine [1]?  In any case, I think we do have a couple of options to consider (java agent should have access to application classpath).

[1] https://github.com/opensearch-project/OpenSearch/pull/18085

cwperks:
> Have you seen the failures related to that?

I haven't tried with JDK24 yet, but I can update this repo to run the same tests with the change in this PR.

This was the run where I saw issues reading classpath resources after https://github.com/opensearch-project/OpenSearch/pull/17989 was merged into core: https://github.com/opensearch-project/security-analytics/actions/runs/14935397717/job/41961528538

sbcd90 (approved):

AWSHurneyt (approved):

PR #1545, opened 2025-06-20
fix correlation engine trigger even if no rules are present

eirsep:
Helping to merge @sbcd90's https://github.com/opensearch-project/security-analytics/pull/1538/files and fix CI


src/main/java/org/opensearch/securityanalytics/transport/TransportCorrelateFindingAction.java
```diff
@@ -137,65 +138,109 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Subscr
...
-                try {
-                    this.correlationIndices.initCorrelationIndex(ActionListener.wrap(response -> {
-                        if (response.isAcknowledged()) {
-                            IndexUtils.correlationIndexUpdated();
-                            if (IndexUtils.correlationIndexUpdated) {
-                                IndexUtils.lastUpdatedCorrelationHistoryIndex = IndexUtils.getIndexNameWithAlias(
-                                        clusterService.state(),
-                                        CorrelationIndices.CORRELATION_HISTORY_WRITE_INDEX
-                                );
+                client.search(searchRequest,
```
engechas, line 157: Making a search call for this every time may get expensive. Is there a way we could cache the results to avoid the check every time this action is called?

src/test/java/org/opensearch/securityanalytics/mapper/MapperRestApiIT.java
```diff
@@ -715,81 +715,6 @@ public void testCreateMappings_withIndexPattern_existing_indexTemplate_update_su
         assertTrue(props.containsKey("destination.port"));
     }
 
-    public void testCreateMappings_withIndexPattern_differentMappings_indexTemplateCleanup_success() throws IOException, InterruptedException {
```
engechas, line 718: Why is this deleted?

sbcd90 (dismissed):

sbcd90 (approved):

riysaxen-amzn (approved):

PR #1648, opened 2026-02-26
Fix bug when deleting detector with 0 rules.

AWSHurneyt:
### Description
The DeleteDetector API is failing with a `groupSize must be greater than 0 but was 0` error when the [[monitorIds]](https://github.com/opensearch-project/security-analytics/blob/main/src/main/java/org/opensearch/securityanalytics/model/Detector.java#L101) attribute of the detector is empty. That error looks to be caused by [[this line]](https://github.com/opensearch-project/security-analytics/blob/main/src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java#L229) on which we pass the list of monitorIds for the detector to the GroupedActionListener we use to listen for DeleteMonitorResponses.

The `monitor_id` attribute becomes an empty array when the detector is created/edited to have 0 rules because without any rules, there are no queries the detector needs to execute; and therefore no monitor is needed for the detector.

I tested this behavior by:
1. Creating a detector with multiple rules associated with it. The detector had a `monitor_id` list with 1 entry.
1. Editing the detector to have only 1 rule. The detector continued to have 1 entry in the list.
1. Edited the detector to have 0 rules. The `monitor_id` list was then empty.

This PR adds a check after [[this line]](https://github.com/opensearch-project/security-analytics/blob/main/src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java#L199) to only perform the monitorDelete API calls if the monitorIds list is not empty.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java
+++ b/src/main/java/org/opensearch/securityanalytics/transport/TransportDeleteDetectorAction.java
@@ -197,6 +197,13 @@ public class TransportDeleteDetectorAction extends HandledTransportAction<Delete
             deleteWorkflow(detector, onDeleteWorkflowStep);
             onDeleteWorkflowStep.whenComplete(acknowledgedResponse -> {
                 List<String> monitorIds = detector.getMonitorIds();
+
+                // A detector with 0 rules will have 0 monitors to delete. Skipping monitor deletion steps.
+                if (monitorIds.isEmpty()) {
+                    deleteDetectorFromConfig(detector.getId(), request.getRefreshPolicy());
+                    return;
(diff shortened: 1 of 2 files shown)


AWSHurneyt:
Looks like the `JDK 21 on macos-latest` and `JDK 25 on macos-latest` checks were failing in this recent PR as well. Will need to investigate why that environment is failing.
https://github.com/opensearch-project/security-analytics/pull/1638/checks

`./gradlew build` (which also runs integration tests) is succeeding locally.
```
BUILD SUCCESSFUL in 35m 58s
30 actionable tasks: 30 executed
```

Issue #1721, opened 2026-05-18
[BUG] [alerting_exception] kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap


NLoi7:
**What is the bug?**
Editing any existing detector via OpenSearch Dashboards (Security Analytics → Detectors → Edit) 
fails with the following error:

[alerting_exception] kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap

The detector is not saved and no changes are applied.

**How can one reproduce the bug?**
1. Go to OpenSearch Dashboards → Security Analytics → Detectors
2. Click on any existing detector
3. Click **Edit**
4. Modify any field (indices, rules, trigger severity, etc.)
5. Click **Save changes**
6. See error: `[alerting_exception] kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap`

**What is the expected behavior?**
The detector should be updated successfully and changes should be reflected immediately.

**What is your host/environment?**
- OS: Ubuntu Server (Docker)
- Version: OpenSearch 3.5.0
- Plugins: opensearch-security-analytics 3.5.0.0, opensearch-alerting 3.5.0.0

**Do you have any additional context?**
- The bug affects **all detectors** regardless of type (windows, network, etc.)
- **Workaround:** Updating detectors via REST API works correctly:
  `PUT /_plugins/_security_analytics/detectors/{detector_id}`
  with the full detector body bypasses the issue entirely.
- This suggests the Dashboard is sending an empty or missing field 
  (likely `ui_metadata`) that causes the plugin to cast `Collections.EmptyMap` 
  to `MutableMap` internally.
- Related issues: #975 (open since April 2024), #1553, #1715

Issue #1553, opened 2025-06-20
[BUG] Attempt to disable an active rule under a threat detector causes an error


tomyerex:
### What is the bug?
Attempt to disable an active rule under a threat detector causes an error: "[alerting_exception] kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap"

### How can one reproduce the bug?
Under "Threat detectors", "Active rules",  an attempt to "Edit" and disable "Suspicious DNS Z Flag Bit Set", and "Save changes" causes an error "Failed to update detector:" -> "[alerting_exception] kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap"

### What is the expected behavior?
The active rule should be updated as disabled.

### What is your host/environment?
Red Hat Enterprise Linux release 8.10 (Ootpa)
Opensearch 3.0.0 (dc4efa821904cc2d7ea7ef61c0f577d3fc0d8be9)
Opensearch Dashboards 3.0.0

### Do you have any screenshots?
![Image](https://github.com/user-attachments/assets/d95c7662-3de5-46c1-b328-d42771fd946b)
![Image](https://github.com/user-attachments/assets/c2022e58-6c26-40c8-a9ea-365c3179955a)

### Do you have any additional context?
_Add any other context about the problem._

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

@AWSHurneyt Can a maintainer take a quick look at this?

AWSHurneyt:
Tagging @Murali-Arigala to assign engineer.

PR #1722, opened 2026-05-18
Fix mutable script params for detector trigger actions (#1715)

### LINKED ISSUE #1715: [BUG] Detector creation/update fails with Kotlin casting error: Collections.EmptyMap cannot be cast to MutableMap
**What is the bug?**
When creating or updating a detector via the Security Analytics API, the operation fails with a 500 Internal Server Error due to a Kotlin casting exception. The error occurs because the DetectorTrigger class uses Java's immutable `Collections.emptyMap()` when creating Script objects for trigger actions, but the underlying Alerting plugin (written in Kotlin) expects a mutable map that can be cast to `kotlin.collections.MutableMap`.

Assumption: Created a detector via the GUI with a trigger in `triggers` but no action in `actions`.
1. Send a PUT request to `/_plugins/_security_analytics/detectors/{detector_id}`
2. Use a detector configuration with triggers that have empty actions arrays: `"actions": []`
3. Include detector triggers in the payload (even with dummy actions, the error persists)
4. Observe the 500 error response with the Kotlin casting exception
Example failing request:

```bash
PUT /_plugins/_security_analytics/detectors/{ID}
{
  "type": "detector",
  "name": "Detector (1h)",
  "detector_type": "apache",
  "enabled": true,
  "schedule": { "period": { "interval": 1, "unit": "HOURS" } },
  "inputs": [{ 
    "detector_input": {
      "description": "Detects token events",
      "indices": ["{index}"],
      "custom_rules": [{"id": "some-id"}]
    }
  }],
  "triggers": [{
    "id": "sjLnr50BoPNAUf777kjD",
    "name": "Token created",
    "severity": "2",
    "types": [],
    "ids": ["hhffr50BZ7K0959c976c"],
    "sev_levels": [],
    "tags": [],
    "actions": [],
    "detection_types": ["rules"]
  }]
}
```

The detector should be successfully created/updated without any casting errors. The system should handle trigger actions gracefully, whether they are empty or contain actual notification actions.

 - OS: Linux
 - Version 3.5.0
 - Plugins
 -- OpenSearch Security Analytics Plugin
 -- OpenSearch Alerting Plugin (Kotlin-based)

Error response:

```
{
  "error": {
    "root_cause": [
      {
        "type": "alerting_exception",
        "reason": "kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap"
      }
    ],
    "type": "security_analytics_exception", 
    "reason": "kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap",
    "caused_by": {
      "type": "alerting_exception",
      "reason": "kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap",
      "caused_by": {
        "type": "exception",
        "reason": "org.opensearch.commons.alerting.util.AlertingException: kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap"
      }
    }
  },
  "status": 500
}
```
- __Root Cause__: Located in `src/main/java/org/opensearch/securityanalytics/model/DetectorTrigger.java` at lines 480-481 in the `getActions()` method

- __Technical Details__: The issue occurs when creating `Script` objects with `Collections.emptyMap()` as parameters. This returns an immutable map that cannot be cast to Kotlin's `MutableMap` interface required by the alerting plugin.

- __Code Location__:

  ```java
  new Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, subjectTemplate, Collections.emptyMap()),
  new Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, messageTemplate, Collections.emptyMap()),
  ```

- __Suggested Fix__: Replace `Collections.emptyMap()` with `new HashMap<>()` to provide mutable empty maps

- __Impact__: This affects all detector operations that involve triggers, preventing users from creating or updating detectors with trigger configurations

thecodingshrimp:
### Description
This change fixes a detector create/update failure caused by using Java's immutable `Collections.emptyMap()` when constructing trigger action `Script` objects in `DetectorTrigger#getActions()`.

The Alerting plugin consumes these script params from Kotlin and expects a mutable map implementation. Replacing the immutable empty map with `new HashMap<>()` prevents the `kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap` exception and allows detector operations with trigger actions to succeed.

### Related Issues
Resolves #1715, #1553, #1721

### Check List
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/model/DetectorTrigger.java
+++ b/src/main/java/org/opensearch/securityanalytics/model/DetectorTrigger.java
@@ -23,6 +23,7 @@ import org.opensearch.script.ScriptType;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -477,8 +478,8 @@ public class DetectorTrigger implements Writeable, ToXContentObject {
                 messageTemplate = messageTemplate.replace("{{ctx.detector", "{{ctx.monitor");
 
                 Action transformedAction = new Action(action.getName(), action.getDestinationId(),
-                        new Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, subjectTemplate, Collections.emptyMap()),
-                        new Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, messageTemplate, Collections.emptyMap()),
+                        new Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, subjectTemplate, new HashMap<>()),
+                        new Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, messageTemplate, new HashMap<>()),
                         action.getThrottleEnabled(), action.getThrottle(),
                         action.getId(), action.getActionExecutionPolicy());
 


thecodingshrimp:
@andrross thx for encouraging

andrross:
@AWSHurneyt Can a maintainer review this PR?

thecodingshrimp:
@AWSHurneyt did you have a chance to look at this or check with a maintainer?

thecodingshrimp:
@toepkerd could you have a look?

andrross:
@thecodingshrimp This seems fine to me, but is it possible for the alerting plugin not to assume anything about the implementation and just make a defensive copy into a mutable map?

AWSHurneyt (approved):

andrross (approved):

cwperks (approved):

thecodingshrimp:
@andrross I actually discovered the real bug while thinking about your comment. I think the real root cause might not be in the alerting plugin at all. The `ClassCastException` (`kotlin.collections.EmptyMap cannot be cast to kotlin.collections.MutableMap`) traces back to `Monitor.suppressWarning()` in `common-utils` repo, which does an unchecked cast `map as MutableMap<String, Any>` on the `uiMetadata` field during `StreamInput` deserialization [reference](https://github.com/opensearch-project/common-utils/blob/50ba24406190b4a6e5d74dcc28debe68a39a2613/src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt#L134). On the SA side, `TransportIndexDetectorAction` passes `Map.of()` (an empty immutable Java map) as `uiMetadata` when constructing the `Monitor` — that gets serialized over transport as a zero-size map, and `StreamInput.readHashMap()` returns `Collections.emptyMap()` for size-0 maps, which then fails the cast. So the fix in this PR addresses the `Script.params` side, but the deeper issue might be that `suppressWarning()` in common-utils is unsafe against any empty immutable map coming through transport. The proper root-cause fix would be in `Monitor.kt`: replacing the unchecked cast with `sin.readMap()?.toMutableMap() ?: mutableMapOf()`. Wdyt and should we issue/PR against common-utils to address this? Still need to do some testing though. Happy to open that if it makes sense.

andrross:
@thecodingshrimp Yeah, it looks like the bug is in common-utils. That Monitor.kt class is using `StreamOutput.writeMap` to write the map, and then `StreamInput.readMap` to read it. It should not make any assumption about the type of the map that survives the serialization round trip other than it implements the `java.util.Map` interface that those methods take. In fact the java doc for readMap() even states the following:

> If the returned map contains any entries it will be mutable. If it is empty it might be immutable.

thecodingshrimp:
@andrross, created a PR and issue in common-utils for this topic.

@AWSHurneyt  What would be the procedure of merging this PR? Can i support the merge in any way?

PR #1735, opened 2026-06-20
Onboards security-analytics plugin to centralized resource authz

### LINKED ISSUE #1747: Onboard security-analytics plugin to Centralized Resource AuthZ framework
Security plugin recently introduced a centralized framework for managing access to resources defined by plugin. This effort centralized the authorization piece that was missing, which led to plugins, like reporting, implementing it within the plugin.

This task aims to onboard security-analytics plugin to centralized authz framework for resources.

- Related to https://github.com/opensearch-project/security/issues/4500

DarshitChanpura:
### Description
Implements resource-access-control for detector and correlation-rule.

### Related Issues
Resolves #1747 

### Check List


DarshitChanpura:
CI will be unblocked once #1736 is merged and this branch gets those changes

DarshitChanpura:
CI failures are unrelated to this PR — all builds fail with `Could not find org.opensearch.plugin:alerting:3.8.0.0-SNAPSHOT`. The alerting plugin snapshot hasn't been published to Maven yet after the version bump to 3.8.0. Will resolve once the upstream artifact is available.

DarshitChanpura:
CI blocked by #1749

DarshitChanpura:
### Code-Diff-Analyzer findings — response (updated for commit `62855a6f`)

I reviewed all 6 findings from the analyzer report. One was a real (low-severity) fix and is included; the rest are correct-by-design or a policy flag on a legitimate first-party dependency. Details below.

| # | Finding | Severity | Disposition |
|---|---------|----------|-------------|
| 1 | New `opensearch-security-spi` compile-only dependency | high | **Needs maintainer bypass** — see below |
| 5 | `ResourceSharingClientAccessor` singleton data race | low | **Fixed** (double-checked locking + `volatile`) |
| 2 | Auth-user header re-injection (`TransportIndexDetectorAction`) | medium | Correct & safe — no change |
| 3 | Auth-user header re-injection (`TransportIndexCorrelationRuleAction`) | medium | Correct & safe — no change |
| 4 | `TransportGetDetectorAction` skips legacy checks when RS active | medium | Intended design — no change |
| 6 | `Detector.docParse` skips unknown top-level fields | low | Necessary & acceptable — no change |

> Note: the branch was squashed/rebased since my earlier reply, so the previously-referenced commit SHA no longer exists. All the above still holds against the current head `62855a6f`; finding #5's fix is present in that tree.

**#1 (high) — `opensearch-security-spi` dependency:** This is the analyzer's mandatory "flag all dependency changes for provenance review" policy, not a code defect. `org.opensearch:opensearch-security-spi` is a first-party OpenSearch artifact pinned to the same `${opensearch_build}` version as every other core dependency, and it is the foundational SPI this entire PR is built on. This is a hard block that cannot be cleared by a code change — it requires a **maintainer to verify provenance and apply the `skip-diff-analyzer` label, then re-run**.

**#5 (low) — accessor data race [FIXED]:** `getInstance()` used unsynchronized lazy init; racing threads could create competing instances, one of which would never receive `setResourceSharingClient()` and leave `shouldUseResourceAuthz()` permanently `false`. Fixed with double-checked locking + `volatile`, matching the reporting plugin's accessor.

**#2 / #3 (medium) — auth-user header re-injection:** The guard only re-injects the persistent authenticated-user marker that the *original request already carried* (captured from the caller's context), and only when it is absent, on an already-stashed context. Detector/correlation-rule creation hands off to alerting-plugin threads (monitor/workflow creation) that don't carry this marker; without re-injection, `ResourceIndexListener.postIndex` cannot record resource ownership. It restores the true caller's identity that was lost across the thread hand-off — it cannot supply an arbitrary or elevated principal, and the write stays a trusted system-index op (no transient user).

**#4 (medium) — GetDetector authorization:** When resource-sharing is active, authorization is *moved* to the security plugin's `ResourceAccessEvaluator` (via the `DocRequest` interface) plus DLS filtering — the legacy backend-role/ownership checks are replaced, not removed. When the resource-sharing client is null/misconfigured, `shouldUseResourceAuthz()` returns `false` and the legacy checks fully apply (fail-safe). This mirrors the reporting and anomaly-detection onboardings.

**#6 (low) — `docParse` skipping unknown fields:** Required so the parser tolerates the `all_shared_principals` field the security plugin injects into stored detector documents. This is the standard forward-compatible-parsing tradeoff.

---

**Maintainer action needed:** finding #1 is a hard block on a legitimate first-party dependency. Please review and apply the `skip-diff-analyzer` label, then re-run the failed action to unblock.

riysaxen-amzn (approved):

PR #1760, opened 2026-08-05
Fix authorization bypass via stashContext() in multiple transport actions

vikhy-aws:
### Summary
- Fix privilege escalation vulnerabilities where `ThreadContext.stashContext()` drops the caller's security context before accessing caller-controlled indices or executing caller-supplied queries.
- Add pre-flight index permission checks for mapping endpoints that accept arbitrary index names.
- Block cross-index `terms` lookup queries in all search endpoints that execute queries with elevated plugin privileges.
- Add `QueryUtils` utility for recursively detecting terms lookups in query trees.

### Description
The security-analytics plugin uses `stashContext()` to elevate privileges when accessing its own internal indices. However, several transport actions applied this elevation to operations involving **caller-controlled** index names or **caller-supplied** query bodies, allowing users to bypass index-level authorization.

#### Two vulnerability classes were identified and fixed:

**Caller-controlled index name (4 endpoints):**
Users could read/write mappings on arbitrary indices they have no permission on by passing the index name to plugin endpoints that stash context before operating.

Fixed by adding a pre-flight `getMappings()` call as the caller (before stashing) to verify index-level permissions:
- `TransportGetMappingsViewAction` — `GET /_plugins/_security_analytics/mappings/view`
- `TransportGetIndexMappingsAction` — `GET /_plugins/_security_analytics/mappings`
- `TransportCreateIndexMappingsAction` — `POST /_plugins/_security_analytics/mappings`
- `TransportUpdateIndexMappingsAction` — `PUT /_plugins/_security_analytics/mappings`

#### Cross-index terms lookup oracle (5 endpoints):
Users could craft search queries with `terms` lookups referencing unauthorized indices. After `stashContext()`, these lookups executed with admin privileges, creating a boolean oracle that reveals whether specific values exist in the target index.

Fixed by rejecting queries containing `TermsQueryBuilder` with a `termsLookup` before context elevation:
- `TransportSearchDetectorAction` — `POST /_plugins/_security_analytics/detectors/_search`
- `TransportSearchRuleAction` — `POST /_plugins/_security_analytics/rules/_search`
- `TransportSearchCustomLogTypeAction` — `POST /_plugins/_security_analytics/logtype/_search`
- `TransportSearchTIFSourceConfigsAction` — `POST /_plugins/_security_analytics/threat_intel/sources/_search`
- `TransportSearchThreatIntelMonitorAction` — `POST /_plugins/_security_analytics/threat_intel/monitors/_search`

### Related Issues
N/A

### Testing
- [x] `QueryUtilsTests` — 16 unit tests covering all query tree traversal paths (null, match_all, inline terms, terms lookup, bool must/should/filter/mustNot, nested bool, constant_score, boosting, dis_max, nested, deeply nested).
- [x] `SecureMappingsAndSearchRestApiIT` — 11 integration tests verifying unauthorized access is rejected and legitimate operations succeed.
- [x] Manual verification: all 14 vulnerability scenarios return 403 after fix.
      1. `GET mappings/view` — read_access user reads unauthorized index fields → 403
      2. `GET mappings` — read_access user reads unauthorized index mappings → 403
    3. `POST mappings` — read_access user writes mappings to unauthorized index → 403
    4. `PUT mappings` — read_access user updates mappings on unauthorized index → 403
    5. `POST detectors/_search` — read_access user with terms lookup to unauthorized index → 403
    6. `POST correlation/rules/_search` — read_access user with terms lookup → 403
    7. `POST rules/_search` (prepackaged) — read_access user with terms lookup → 403
    8. `POST rules/_search` (custom) — read_access user with terms lookup → 403
    9. `POST mappings` — full_access user (cluster perm only, no index perm) writes to unauthorized index → 403
    10. `PUT mappings` — full_access user (cluster perm only, no index perm) updates unauthorized index → 403
    11. `POST correlation/rules/_search` — full_access user with terms lookup → 403
    12. `POST logtype/_search` — read_access user with terms lookup → 403
    13. `POST threat_intel/sources/_search` — full_access user with terms lookup → 403
    14. `POST threat_intel/monitors/_search` — full_access user with terms lookup → 403
- [x] Regression: admin and authorized users can still use all endpoints normally.
- [x] Run full `./gradlew integTest -Dsecurity=true -Dhttps=true` suite.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/threatIntel/transport/monitor/TransportSearchThreatIntelMonitorAction.java
+++ b/src/main/java/org/opensearch/securityanalytics/threatIntel/transport/monitor/TransportSearchThreatIntelMonitorAction.java
@@ -22,11 +22,13 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.index.IndexNotFoundException;
 import org.opensearch.search.SearchHit;
+import org.opensearch.search.builder.SearchSourceBuilder;
 import org.opensearch.securityanalytics.settings.SecurityAnalyticsSettings;
 import org.opensearch.securityanalytics.threatIntel.action.monitor.SearchThreatIntelMonitorAction;
 import org.opensearch.securityanalytics.threatIntel.action.monitor.request.SearchThreatIntelMonitorRequest;
 import org.opensearch.securityanalytics.threatIntel.sacommons.monitor.ThreatIntelMonitorDto;
 import org.opensearch.securityanalytics.threatIntel.util.ThreatIntelMonitorUtils;
+import org.opensearch.securityanalytics.transport.QueryUtils;
 import org.opensearch.securityanalytics.transport.SecureTransportAction;
 import org.opensearch.tasks.Task;
 import org.opensearch.threadpool.ThreadPool;
@@ -77,6 +79,14 @@ public class TransportSearchThreatIntelMonitorAction extends HandledTransportAct
             listener.onFailure(new OpenSearchStatusException("Do not have permissions to resource", RestStatus.FORBIDDEN));
             return;
         }
+
+        SearchSourceBuilder source = request.searchRequest().source();
+        if (source != null && source.query() != null && QueryUtils.containsTermsLookup(source.query())) {
+            listener.onFailure(new OpenSearchStatusException(
+                    "Terms lookup queries referencing external indices are not permitted", RestStatus.FORBIDDEN));
+            return;
+        }
+
         this.threadPool.getThreadContext().stashContext();
 
         //TODO change search request to fetch threat intel monitors
--- a/src/main/java/org/opensearch/securityanalytics/threatIntel/transport/TransportSearchTIFSourceConfigsAction.java
+++ b/src/main/java/org/opensearch/securityanalytics/threatIntel/transport/TransportSearchTIFSourceConfigsAction.java
@@ -13,7 +13,9 @@ import org.opensearch.common.settings.Settings;
 import org.opensearch.commons.authuser.User;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.rest.RestStatus;
+import org.opensearch.search.builder.SearchSourceBuilder;
 import org.opensearch.securityanalytics.settings.SecurityAnalyticsSettings;
+import org.opensearch.securityanalytics.transport.QueryUtils;
 import org.opensearch.securityanalytics.threatIntel.action.SASearchTIFSourceConfigsAction;
 import org.opensearch.securityanalytics.threatIntel.action.SASearchTIFSourceConfigsRequest;
 import org.opensearch.securityanalytics.threatIntel.service.DefaultTifSourceConfigLoaderService;
@@ -66,6 +68,14 @@ public class TransportSearchTIFSourceConfigsAction extends HandledTransportActio
             actionListener.onFailure(new OpenSearchStatusException("Do not have permissions to resource", RestStatus.FORBIDDEN));
             return;
         }
+
+        SearchSourceBuilder source = request.getSearchSourceBuilder();
+        if (source != null && source.query() != null && QueryUtils.containsTermsLookup(source.query())) {
+            actionListener.onFailure(new OpenSearchStatusException(
+                    "Terms lookup queries referencing external indices are not permitted", RestStatus.FORBIDDEN));
+            return;
+        }
+
         this.threadPool.getThreadContext().stashContext(); // stash context to make calls as admin client
 
         StepListener<Void> defaultTifConfigsLoadedListener;
--- a/src/main/java/org/opensearch/securityanalytics/transport/QueryUtils.java
+++ b/src/main/java/org/opensearch/securityanalytics/transport/QueryUtils.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.securityanalytics.transport;
+
+import org.apache.lucene.search.BooleanClause;
+import org.opensearch.index.query.QueryBuilder;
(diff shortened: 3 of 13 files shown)


riysaxen-amzn:
Thanks for tackling this — the two vulnerability classes are real and the fix direction is right. A few items before I can approve:

🔴 Blocking: Pre-flight check uses wrong permission level (Create/Update mappings)
TransportCreateIndexMappingsAction and TransportUpdateIndexMappingsAction use GetMappingsRequest (requires indices:admin/mappings/get) as the authorization probe, but the actual operation performs putMapping (requires indices:admin/mapping/put). A user with read-only index access passes the pre-flight check and then writes mappings via the stashed admin context.

The bot reviewer caught this too. Fix: use PutMappingRequest (or equivalent write-level probe) for the create/update actions. The GET mappings actions are fine with GetMappingsRequest.

🟡 Should-fix: QueryUtils.containsTermsLookup incomplete traversal
The traversal handles: bool, constant_score, boosting, dis_max, nested, function_score. It misses:

has_child / has_parent (join module)
script_score
span_* wrappers (unlikely but possible)
An attacker could wrap terms { lookup: {...} } inside has_child to bypass. Suggested fix: either add the missing types, or flip to a deny-by-default approach — use QueryBuilder.visit() / rewrite() to walk the full tree generically, or reject any query containing an unknown wrapper type.

🟡 Should-fix: Redundant private containsTermsLookup in TransportSearchDetectorAction
The private method first checks instanceof TermsQueryBuilder and returns early, then falls through to QueryUtils.containsTermsLookup() which handles the same case. This is dead code — just delegate directly to QueryUtils.

🟠 Nit: Commented-out filter-by-user block in TransportSearchThreatIntelMonitorAction
Lines 76 and 92 have TODO comments about filter-by-user being "currently commented out." If this is intentional/pre-existing, it's fine for this PR. But it should be called out that this endpoint has an additional pre-existing gap — unauthenticated users may see other users' monitors.

🟠 Nit: GetMappingsViewAction may still leak sample documents
The bot noted that getMappingsViewAction() (called after stashContext in the view endpoint) may read sample documents from the index. If the pre-flight only checks mapping metadata access, a user without indices:data/read/search could still trigger sample doc reads via elevated privileges. Worth a follow-up.

src/main/java/org/opensearch/securityanalytics/threatIntel/transport/monitor/TransportSearchThreatIntelMonitorAction.java
```diff
@@ -77,6 +79,14 @@ protected void doExecute(Task task, SearchThreatIntelMonitorRequest request, Act
             listener.onFailure(new OpenSearchStatusException("Do not have permissions to resource", RestStatus.FORBIDDEN));
             return;
         }
+
```
eirsep, line 82: TransportSearchCorrelationRuleAction  has the identical bug and is NOT patched

 RestSearchCorrelationRuleAction builds the query from the caller's request body (searchSourceBuilder.parseXContent(request.contentOrSourceParamParser())) — identical to the
  patched RestSearchRuleAction.
  - TransportSearchCorrelationRuleAction.doExecute then calls stashContext() and passes request.getSearchRequest() straight to client.search() with no containsTermsLookup check.

eirsep (changes requested):
Another thing is that 
`containsTermsLookup` is fail-open — incomplete traversal can be bypassed (security flaw)

  This method is the enforcement point for the terms-lookup block, but it's built as a hardcoded allowlist of wrapper types (bool, constant_score, boosting, dis_max, nested,
  function_score) with a trailing return false. Anything it doesn't recognize is treated as **safe**. For a security gate, "unrecognized → allowed" is the wrong default — a caller only has to wrap the termsLookup in a query type this ladder doesn't enumerate to slip past it.

  Two concrete gaps I've confirmed:

  1. function_score filter functions are not traversed. The code checks FunctionScoreQueryBuilder.query(), but FunctionScoreQueryBuilder also exposes filterFunctionBuilders(), and each FilterFunctionBuilder carries its own filter query (getFilter()) that can hold a termsLookup. That path is never visited, so this bypasses the check today:

```
  {
    "query": {
      "function_score": {
        "functions": [
          {
            "filter": {
              "terms": {
                "detector.name": { "index": "hr-salaries-test", "id": "1", "path": "employee_national_id" }
              }
            },
            "weight": 1
          }
        ]
      }
    }
  }
```
  2. Other wrapper types aren't covered at all — e.g. `script_score, has_child/has_parent, and any future query type.` Each is another silent bypass.

  Rather than extending the instanceof ladder (which just moves the goalposts — the next unlisted wrapper reopens the hole), I'd recommend traversing with the framework's own  visitor, QueryBuilder.visit(**QueryBuilderVisitor**), which recurses into all registered sub-queries. That makes the check complete by construction instead of by enumeration, so it  can't be walked around by wrapping the payload.

riysaxen-amzn:
Thanks, the updated revision addresses my concerns:

- Write-level `PutMappingRequest` dry-run via `perform_permission_check` correctly closes the read-probe gap on create/update mappings, with a proper fallback when the security plugin isn't installed.
- Switching `QueryUtils` to `QueryBuilderVisitor` gives deny-by-default traversal coverage (has_child/has_parent/script_score and future wrappers included).
- Correlation rule search now has the guard, and the redundant private helper is gone.

Approving. Two small asks: (1) please confirm the `-Dsecurity=true` integ suite passed on this revision (the FORBIDDEN detection relies on the dry-run response shape), and (2) consider a follow-up for `getMappingsViewAction` sample-document reads, which still execute under the stashed context after a mappings-metadata-only permission check.

riysaxen-amzn (approved):

vikhy-aws:
Removed `perform_permission_check` which caused integ tests to fail as it can be passed on on REST calls and not on transport call. 

**Note:** The only trade-off with this is that cluster state version is bumped on successful permission check which should not be an issue.

vikhy-aws:
All integ tests passed.
Screenshot 2026-08-05 at 5 44 26 PM

eirsep (approved):

PR #1789, opened 2026-08-14
fix: Prevent findings history deletion by alert retention period

### LINKED ISSUE #1759: [BUG] Findings history deleted by alert retention instead of finding retention
### Summary

Findings history indices are deleted based on `alert_history_retention_period`, not `finding_history_retention_period`.

In `DetectorIndexManagementService`, cleanup applies alert, finding, correlation, and IOC retention checks to every scanned index. A findings index has no alerts-history alias, so once it is older than **alert** retention it is marked for delete—even when finding retention is longer.

### Steps to reproduce

1. Create a detector that generates findings.
2. Apply these cluster settings:

```json
PUT _cluster/settings
{
  "persistent": {
    "plugins.security_analytics.finding_history_max_age": "2d",
    "plugins.security_analytics.finding_history_retention_period": "4d",
    "plugins.security_analytics.alert_history_retention_period": "2d"
  }
}
```

3. Generate findings and wait for the findings index to roll over after `finding_history_max_age` (2 days).
4. Wait until the rolled findings index is older than `alert_history_retention_period` (2 days), but still younger than `finding_history_retention_period` (4 days).
5. Check indices after the next cleanup job:

```text
GET _cat/indices/.opensearch-sap-*-findings*?v
```

### Example

1. Day 0 — findings index created
2. Day 2 — index rolls over (`max_age`)
3. Day 2+ — old findings index is deleted because age > **alert** retention (2d)

Result: findings are gone after ~2 days, not 4.

### Expected behavior

- Findings indices are retained for `finding_history_retention_period`
- Alert history indices are retained for `alert_history_retention_period`
- Each history type uses only its own retention setting

### Reference

For reference, this was the file I checked: `src/main/java/org/opensearch/securityanalytics/indexmanagment/DetectorIndexManagementService.java`.

nagendramohan:
The getHistoryIndexToDelete method incorrectly returned the index name for deletion when the index had no matching alias in the history type being checked. A findings index with no alert-history alias would be deleted after exceeding alert_history_retention_period instead of finding_history_retention_period.

Fix: return null (do not delete) when an index does not belong to the history type being evaluated (alias == null). Only delete when the index positively matches the history type's alias AND exceeds that type's retention period.

Also makes getHistoryIndexToDelete package-private static for testability and adds unit tests covering:
- Findings index not deleted by alert retention check
- Findings index deleted when exceeding finding retention
- Findings index not deleted when younger than retention

Resolves opensearch-project/security-analytics#1759

### Description
  Fixes the getHistoryIndexToDelete method in DetectorIndexManagementService which incorrectly deleted findings indices based on alert_history_retention_period instead of finding_history_retention_period.
  
  Root cause: The method is called once per history type (alert, finding, correlation, IOC) for each index. When an index had no matching alias for the history type being checked (e.g., a findings index checked against alert history aliases), the method fell through and returned the index for deletion. It should have returned null (skip — this index doesn't belong to me).
  
  Fix: Return null when no alias match is found, meaning the index doesn't belong to this history type. Only return the index name for deletion when it positively matches the type's alias AND exceeds that type's retention period.
  
  Also makes getHistoryIndexToDelete package-private static (uses no instance state) and adds 3 unit tests.

### Related Issues
Resolves #1759

### Check List
- [x] New functionality has been documented.

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/indexmanagment/DetectorIndexManagementService.java
+++ b/src/main/java/org/opensearch/securityanalytics/indexmanagment/DetectorIndexManagementService.java
@@ -384,7 +384,8 @@ public class DetectorIndexManagementService extends AbstractLifecycleComponent i
         return indicesToDelete;
     }
 
-    private String getHistoryIndexToDelete(
+    // Package-private and static for testability; does not use instance state.
+    static String getHistoryIndexToDelete(
             IndexMetadata indexMetadata,
             Long retentionPeriodMillis,
             List<HistoryIndexInfo> historyIndices,
@@ -408,8 +409,11 @@ public class DetectorIndexManagementService extends AbstractLifecycleComponent i
                     // If the index has the write alias and history is enabled, don't delete the index
                     return null;
                 }
+                // Index belongs to this history type but history is disabled — delete it
+                return indexMetadata.getIndex().getName();
             }
-            return indexMetadata.getIndex().getName();
+            // Index does not belong to this history type — do not delete
+            return null;
         }
         return null;
     }
--- a/src/test/java/org/opensearch/securityanalytics/indexmanagment/DetectorIndexManagementServiceTests.java
+++ b/src/test/java/org/opensearch/securityanalytics/indexmanagment/DetectorIndexManagementServiceTests.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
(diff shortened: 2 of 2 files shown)


nagendramohan:
Gentle bump 🙂 — this fixes a retention bug where findings-history indices could be deleted on the alert retention period (and vice-versa) because getHistoryIndexToDelete didn't check the index belongs to the history type being processed, causing data to be removed earlier than configured. Small,  tested change. Could a maintainer take a look when there's bandwidth? Happy to address feedback. Thanks!

PR #1804, opened 2026-08-28
fix: replace _ws_ space encoding with escaped spaces in SIGMA query generation

### LINKED ISSUE #1024: SIGMA rule translation -> lucene query replaces spaces " "  with "_ws_" which lucene doesnt understand.
**What is the bug?**
A sigma rules with a whitespace gets incorrectly translated to a search query with "_ws_" instead of a whitespace.

detection:
  condition: (selection and selection_evt) and not filter_main_known_locations
  filter_main_known_locations:
    process.executable|contains:
    - C:\\Program Files (x86)\\Windows Defender\\
    - C:\\Program Files\\Microsoft Security Client\\
    - C:\\Program Files\\Windows Defender\\
    - C:\\ProgramData\\Microsoft\\Windows Defender\\Platform\\
    - C:\\Windows\\WinSxS\\
  selection:
    process.name:
    - MpCmdRun.exe
    - NisSrv.exe
  selection_evt:
    event.code: 1
    event.module: sysmon

**Results in:**
	(((process.name: "MpCmdRun.exe") OR (process.name: "NisSrv.exe")) AND ((event.code: 1) AND (event.module: "sysmon"))) AND ((((NOT process.executable: *C\:\\Program_ws_Files_ws_\(x86\)\\Windows_ws_Defender\\* AND _exists_: process.executable) AND (NOT process.executable: *C\:\\Program_ws_Files\\Microsoft_ws_Security_ws_Client\\* AND _exists_: process.executable) AND (NOT process.executable: *C\:\\Program_ws_Files\\Windows_ws_Defender\\* AND _exists_: process.executable) AND (NOT process.executable: *C\:\\ProgramData\\Microsoft\\Windows_ws_Defender\\Platform\\* AND _exists_: process.executable) AND (NOT process.executable: *C\:\\Windows\\WinSxS\\* AND _exists_: process.executable))))

A whitespace should work properly in a sigma rule that is translated to a lucene query

OS 2.13

Looking in github, I can see where the "_ws_" is being set. Not sure why.
https://github.com/search?q=repo%3Aopensearch-project%2Fsecurity-analytics%20%20_ws_&type=code

thecodingshrimp:
### Description

SIGMA detection rules whose field values contain spaces never match because spaces are encoded as literal `_ws_` tokens that Lucene cannot decode in all query contexts. This fix replaces the `_ws_` scheme with backslash-escaped spaces (`\ `) emitted via the existing backend escape loop.

**Two failure modes fixed:**

1. **Wildcard path** (`contains`/`startswith`/`endswith`): `query_string` wildcard terms bypass the field analyzer, so `_ws_` is matched literally against ingest data that contains real spaces — detections never fire.
2. **Regex path**: `SigmaRegularExpression` encoded spaces as `_ws_`, but Lucene regex syntax treats space as a literal character and does not apply the `rule_ws_filter` char_filter — detections never fire.

**Additional correctness fix:**

3. `SigmaWindowsDashModifier` hard-coded a `.replace("_ws_", " ")` decode followed by `.replace(" ", "_ws_")` re-encode. This conflicted with the new escaping on the wildcard path and has been removed.

**Changes:**

- `OSQueryBackend.java`: add `" "` (space) to `addEscaped` so the escape loop in `SigmaString.convert()` emits `\ ` via backend config rather than hardcoded string replace.
- `SigmaString.java`: remove hardcoded `replace(" ", "_ws_")` from `convert()` and `toString()` — escaping now flows through the backend-config escape loop.
- `SigmaRegularExpression.java`: remove `replace(" ", "_ws_")` — Lucene regex treats space as a literal, no encoding needed.
- `SigmaWindowsDashModifier.java`: remove `_ws_` decode/re-encode round-trip — values are stored with literal spaces throughout.
- `detector-settings.json` / `DetectorMonitorConfig.java`: `rule_ws_filter` and `rule_analyzer` are kept during this transition window; the `_ws_` token no longer appears in generated queries so the filter is dormant but harmless. Retirement requires a re-index migration and is deferred to a follow-up.

### Related Issues

Resolves #1024

### Check List

  - Unit tests: `QueryBackendTests` — whitespace wildcard path (`*This\ is\ an\ example*`), whitespace quoted path, regex path, windash path; all passing (`BUILD SUCCESSFUL`).
  - Integration tests: `DetectorRestApiIT.testDetectorWithWhitespaceRuleContainsPath` and `testDetectorWithWhitespaceRuleQuotedPath` added (compile-verified; require a live OpenSearch cluster to run).
- [ ] API changes companion pull request created.
- [x] Public documentation issue/PR created.

### DIFF
--- a/src/main/java/org/opensearch/securityanalytics/config/monitors/DetectorMonitorConfig.java
+++ b/src/main/java/org/opensearch/securityanalytics/config/monitors/DetectorMonitorConfig.java
@@ -61,6 +61,8 @@ public class DetectorMonitorConfig {
 
     public static Map<String, Map<String, String>> getRuleIndexMappingsByType() {
         HashMap<String, String> properties = new HashMap<>();
+        // rule_ws_filter retained for backward compat with detectors compiled before PR #1789.
+        // Do not remove until all persisted detectors have been re-saved.
         properties.put("analyzer", "rule_analyzer");
         HashMap<String, Map<String, String>> fieldMappingProperties = new HashMap<>();
         fieldMappingProperties.put("text", properties);
--- a/src/main/java/org/opensearch/securityanalytics/rules/backend/OSQueryBackend.java
+++ b/src/main/java/org/opensearch/securityanalytics/rules/backend/OSQueryBackend.java
@@ -34,6 +34,7 @@ import org.opensearch.securityanalytics.rules.types.SigmaCIDRExpression;
 import org.opensearch.securityanalytics.rules.types.SigmaCompareExpression;
 import org.opensearch.securityanalytics.rules.types.SigmaExpansion;
 import org.opensearch.securityanalytics.rules.types.SigmaNumber;
+import org.opensearch.securityanalytics.rules.types.Placeholder;
 import org.opensearch.securityanalytics.rules.types.SigmaRegularExpression;
 import org.opensearch.securityanalytics.rules.types.SigmaString;
 import org.opensearch.securityanalytics.rules.utils.AnyOneOf;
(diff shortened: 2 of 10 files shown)


thecodingshrimp:
Thanks for the thorough review. Addressing both flagged areas:

**Quoted string escaping (`addEscaped` + space)**

The concern is valid to raise, but `\ ` inside a Lucene `query_string` quoted phrase resolves correctly. The Lucene query parser treats backslash as the escape character in all contexts — both wildcard terms and quoted phrases — so `"This\ is\ an\ example"` is parsed as the phrase `This is an example` and matches ingest data containing plain spaces. This is the same escaping used by other Lucene-based backends (e.g. Elasticsearch/OpenSearch query DSL). The unit test `testConvertValueStrWithWhitespaceQuoted` asserting `mappedA: "This\ is\ an\ example"` reflects the correct emitted form.

**Null regexp on error path**

The assignment ordering is preserved (`this.regexp = regexp` before `compile()`). Regarding null input: `Pattern.compile(null)` throws `NullPointerException`, which is identical behavior to the previous code where `regexp.replace(" ", "_ws_")` would also throw NPE on null. The null contract is unchanged.

cudos: claude

